PR #9, opened 2026-06-18
Feature/install register separation

nhuelstng:
- Unregistered skills in a project do not block starting omac but give a warning
- Command to uninstall skills

-> if you installed a skill by accident (or you currently do not want to set it up but continue working) you are no longer blocked


thomas-endres-tng:
I also implemented something similar. Can you check what is still needed here?

nhuelstng:
I think the manifest functionality is still missing, but maybe it makes sense to redo that and first brainstorm about how we really want to have it. Or is it just overengineered?
Some thoughts:
- if we really want a reproducible setup we need the options to exclude / ignore global skills (apart from exceptions like the marketplace)
- is name as an unique identifier enough? maybe a scope prefix in the asml would be nice like global/marketplace tng/slack. the skill id is currently not visible in the frontend?

thomas-endres-tng:
Let's talk about it once more before merging.

PR #15, opened 2026-06-24
fix(sandbox): grant harness shim dir, not just symlink target

nhuelstng:
## Problem

`omac start` aborts before the harness even launches:

```
omac sandbox stage2: cannot run "opencode" inside the sandbox:
"opencode" not found in the sandbox filesystem.
```

…whenever the inner-harness binary on `PATH` is a **shim/symlink whose target lives in a different directory tree**. The case that exposed it is opencode installed via **bun**:

- `~/.bun/bin/opencode` — the entry on `PATH` (a symlink)
- → `~/.bun/install/global/node_modules/opencode-ai/bin/opencode.exe` — the real binary

## Root cause

`BuildChildArgv` already grants the harness binary's directory into the namespace (introduced in b358c9b *"grant inner harness binary dir so version-manager installs work"*). But `resolveInnerBinaryDir` returned **only the symlink-resolved** directory:

```go
resolved, _ := exec.LookPath(argv[0])      // ~/.bun/bin/opencode        (the shim)
resolved      = filepath.EvalSymlinks(...)  // ~/.bun/.../opencode-ai/bin/opencode.exe
return filepath.Dir(resolved)              // grants ONLY .../opencode-ai/bin
```

So the directory of the shim that is actually **on `PATH`** (`~/.bun/bin`) was never mounted. Inside the namespace, stage2 re-runs `LookPath("opencode")`, walks `PATH` to `~/.bun/bin` — which isn't there — and fails, even though the resolved binary's directory *was* granted.

## Fix

Grant **both** directories:

- the **PATH-entry (shim) dir**, so stage2's own `LookPath` finds the binary on `PATH`;
- the **symlink-resolved dir**, so the link target and its sibling files (shared libs, node runtime) are reachable.

`resolveInnerBinaryDir` → `resolveInnerBinaryDirs` returning `[]string` (deduped when the two coincide, e.g. a non-symlinked binary).

This removes the need for sandbox profiles to hand-enumerate version-manager paths — omac resolves wherever the harness actually lives, for bun, npm-global, brew, and similar layouts.

## Verification

- Updated unit test covers the bun-style **shim + `.exe` rename across trees** and the **non-symlink** (single-dir) case.
- End-to-end: with all `~/.bun/*` entries **removed** from the sandbox profile read-list, a patched omac runs `opencode --version` inside the sandbox successfully (`1.17.9`); the unpatched binary fails with the error above.
- `go build ./...`, `go vet ./internal/sandboxrun/`, and the sandboxrun test suite are clean.

## Known limitation

This covers **symlinked** installs (bun, npm-global, brew). Shims that are executable *scripts* re-exec'ing a version manager at runtime (some mise/asdf setups) resolve to the script itself; making those reachable would require granting the manager binary plus its install tree, which is out of scope here.

### DIFF
--- a/internal/sandboxrun/backend_linux.go
+++ b/internal/sandboxrun/backend_linux.go
@@ -134,30 +134,46 @@ func kernelVersionString() string {
 	return strings.TrimSpace(string(data))
 }
 
-// resolveInnerBinaryDir resolves the inner command's executable on the
-// host PATH and returns the directory holding its real (symlink-resolved)
-// file, or "" when the command cannot be found or resolved. It runs on
-// the supervisor (outside the namespace), so the lookup sees the user's
-// real PATH — the same resolution `which opencode` performs. Directories
-// already covered by the baseline (e.g. /usr/bin) are harmless: the
-// dedupe in BuildBwrapArgv collapses the duplicate grant.
-func resolveInnerBinaryDir(innerArgv []string) string {
+// resolveInnerBinaryDirs resolves the inner command's executable on the
+// host PATH and returns the directories that must be granted for it to be
+// reachable inside the namespace:
+//
+//   - the directory of the PATH entry itself, which is frequently a symlink
+//     or shim (e.g. ~/.bun/bin/opencode, a mise/asdf shim). stage2 re-runs
+//     LookPath inside the namespace, so this dir must be mounted and on PATH
+//     or the lookup fails even when the real binary is present elsewhere.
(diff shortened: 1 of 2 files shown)


nhuelstng:
Rebased onto #18 (netproxy CI flake fix) and retargeted this PR's base to that branch, so CI runs with the fix and the diff here stays limited to the sandbox change. **Merge #18 first**; GitHub will auto-retarget this PR to `main`.

PR #16, opened 2026-06-25
feat(cli): omac continue and resume session commands

nhuelstng:
## What

Two new subcommands that re-enter prior work through omac's existing sandboxed launch pipeline:

- **`omac continue [harness]`** — re-launch the most recent session for the current workdir (appends the harness continue flag: opencode/claude `--continue`).
- **`omac resume [harness]`** — list this workdir's recent sessions in an interactive numbered picker (index, relative time, session id, title) and launch the selected one (opencode `--session <id>`, claude `--resume <id>`).

Both shipped harnesses are first-class.

## How

- **`internal/config`** — harness descriptors gain a `Session` block (`ContinueArgs`, `ResumeByIDArgs`, `ListKind`) — pure data, no I/O.
- **`internal/session`** — new package listing sessions per harness, scoped to the workdir:
  - opencode: parse `opencode session list --format json`, keep records whose `directory` is the workdir.
  - Claude Code: read `~/.claude/projects/<encoded-cwd>/*.jsonl` (id = filename, title = latest `aiTitle` → first user message → id, time = latest record timestamp → mtime). The lossy dir encoding is only a lookup hint — membership is confirmed against each file's embedded `cwd`.
  - Best-effort throughout: a missing CLI/store yields an empty list, never a hard error.
- **`internal/cli/start.go`** — extracted `runLaunch` + `parseLaunchArgs` so `start`, `continue`, and `resume` share one launch implementation (behavior-preserving; existing start tests pass).
- **`internal/cli`** — `continue.go`, `resume.go` (numbered picker, non-TTY fallback, cancel handling), dispatcher + usage entries.

Designed and tracked via OpenSpec: `openspec/changes/omac-continue-resume/` (proposal, design, specs, tasks — all 25 tasks complete).

## Scope notes

- Sessions are scoped to the **exact** current workdir (path match, not prefix; symlinks not resolved).
- **Provenance coloring is intentionally out of scope** (which sessions were run in omac vs. directly) — omac has no authoritative signal short of a harness plugin; deferred to a follow-up. See the proposal's Out of Scope.

## Testing

- `go build ./...`, `go vet`, full `go test ./...` all pass; unit tests for both listing backends, the harness metadata, opts building, selection parsing, and the non-TTY fallback.
- Verified against the real stores on this machine: `omac resume` (opencode) and `omac resume claude` both list the correct workdir-scoped sessions, newest-first.
- Verified `continue` → launch wiring end-to-end (`omac continue --no-sandbox --inner /bin/echo -- hello` runs `echo --continue hello`).
- Not exercised in a real TTY here: the interactive selection keystroke path (piped stdin correctly falls back to the list + hint). The selection logic and launch path are unit-tested, but a hands-on terminal run is worth doing before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,25 @@ Supported harnesses (and aliases): `opencode` (`oc`), `claude-code`
 is rejected with the list of supported names. Inner arguments that happen to be
 barewords go after `--` (e.g. `omac start claude -- --model sonnet`).
 
+### Resuming prior work
+
+Two convenience subcommands re-enter earlier sessions through the same
+sandboxed launch pipeline as `start`:
+
+```bash
+omac continue          # reopen the last session for this folder (opencode)
+omac continue claude   # ...with Claude Code
+omac resume            # pick from this folder's recent sessions, then launch
+omac resume claude     # ...with Claude Code
+```
+
+`omac resume` lists only the current folder's sessions, newest first, and
+launches the one you pick inside omac. It reads each harness's own session
+store — opencode via `opencode session list`, Claude Code by reading
+`~/.claude/projects/<folder>/<id>.jsonl` (titles come from the session's
+`aiTitle`, and a file's embedded `cwd` decides which folder it belongs to).
+Both subcommands take the same flags and optional `[harness]` token as `start`.
+
 Each harness ships a small client-side **bridge** that wires the agent to
 omac's control plane (skill activation, the skills manifest, skill base URLs):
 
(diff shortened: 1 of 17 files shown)


nhuelstng:
Thanks for the review — addressed all five in 3f05a5d:

1. **README Claude path** — now documents `~/.claude/projects/<encoded-cwd>/<id>.jsonl` and explains the non-alphanumeric→`-` encoding, instead of the misleading `<folder>`.
2. **Timestamp parsing** — switched to `time.RFC3339Nano`. (Note: Go's `time.Parse` is actually lenient and already accepted the fractional form with `RFC3339`, so this wasn't a live bug — but `RFC3339Nano` is clearer and removes any doubt.)
3. **Ctrl-C / TTY** — `pickSession` now intercepts `SIGINT` during selection and treats it as cancel (exit OK) instead of relying on the default handler's non-zero kill, and it now requires **both** stdin and stdout to be a TTY before prompting (matches the spec).
4. **Imports** — added `os` and `os/signal` for the above (didn't need `context` with the goroutine+select approach).
5. **`omac start:` prefix** — threaded the invoking subcommand name through `launchOpts` (`label`) and `runLaunch` now prefixes every diagnostic with it, so failures via `omac continue`/`omac resume` are labeled correctly. Verified: `omac continue --sandbox nope` → `omac continue: unknown sandbox profile "nope"`.

Build, `staticcheck ./...`, `gofmt`, and `go test -race ./...` all clean.

PR #17, opened 2026-06-25
fix(start): let env_passthrough satisfy required secrets, validate pattern

nhuelstng:
## Problem

`omac start` refused to launch with `required secret missing` for skills like the skainet set, **even when the secret was exported in the shell and the skill declared it under `env_passthrough`**.

The skill manifests and `creating-a-skill.md` §10.3 document a "keychain **or** `env_passthrough`" contract: a shell that exports the token should keep working in keychain-less environments (CI runners, headless servers). At runtime `supervisor.buildEnv` honors this — it injects `env_passthrough` vars from the host. But the `omac start` **preflight** resolved required secrets from the **keychain only** (`start.go`), so it refused before runtime injection ever happened.

The tell: the config-field preflight right below it *does* honor `default_from_env` from the host env. The secret loop just lacked the equivalent check.

## Fix

1. **Honor the fallback** — a required secret absent from the keychain is no longer reported missing if it's listed under `sidecar.env_passthrough` *and* the shell exports a non-empty value (the exact conditions under which the supervisor injects it at runtime).
2. **Validate shape** — an `env_passthrough`-supplied value is checked against the secret's `pattern` (reusing the register flow's `validatePattern`). Keychain values were vetted at register time; this is the only point an env-supplied one gets vetted before the sidecar receives it. A mismatch is reported as its own problem class rather than a misleading "missing" message.
3. **Escape hatch** — `--skip-secret-pattern` bypasses the pattern check for the case where the declared pattern is outdated. The raw value is still passed through.

## Behavior for a keychain-absent secret

| Situation | Result |
|---|---|
| Not in env_passthrough / env unset / empty | `required secret missing` (unchanged) |
| In env_passthrough, value matches pattern | starts; supervisor injects at runtime |
| In env_passthrough, value present but fails pattern | refuses to start (new class); `--skip-secret-pattern` overrides |

## Tests

- `TestSecretFromEnv` — 4 cases (in passthrough + set; unset; empty; not in passthrough).
- `TestValidatePattern_EnvSuppliedSecret` — match / mismatch / no-pattern, using the skainet `SKAINET_TOKEN` shape.

`go build`, `go vet`, and the full `internal/cli` suite pass. Docs (`README.md`, `oh-my-agentic-coder.md`) updated to list the new flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,8 @@ omac [--workdir <dir>] <subcommand> [flags] [args]
                  --no-sandbox            debug: run inner cmd directly
                  --keep-running          don't stop sidecars on exit
                  --accept-skill-changes  tolerate bundle_hash drift
+                 --skip-secret-pattern   don't enforce a secret's pattern
+                                         on an env_passthrough value
                  --verbose               lifecycle logging
 
   doctor       Sanity checks: config, registry, binaries, secrets, sandbox.
--- a/internal/cli/start.go
+++ b/internal/cli/start.go
@@ -34,6 +34,7 @@ func runStart(args []string, env *Env) int {
 		noSandbox          = fs.Bool("no-sandbox", false, "Run inner command directly, without a sandbox (debug only).")
 		keepRunning        = fs.Bool("keep-running", false, "Do not stop sidecars when the inner command exits.")
 		acceptSkillChanges = fs.Bool("accept-skill-changes", false, "Tolerate bundle_hash drift in registered skills (proceed even if the on-disk skill differs from what was registered).")
+		skipSecretPattern  = fs.Bool("skip-secret-pattern", false, "Do not enforce a secret's pattern against an env_passthrough-supplied value (escape hatch for an outdated pattern; the raw value is still passed through).")
 		verbose            = fs.Bool("verbose", false, "Verbose lifecycle logging.")
 	)
 	fs.Usage = func() {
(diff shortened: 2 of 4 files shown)


nhuelstng:
Thanks — all three review comments addressed in 0dce141:

1. **`secretFromEnv` doc comment (start.go)** — You're right that `supervisor.buildEnv` passes a var through on *presence* even when empty, while this helper requires non-empty. Reworded the comment to stop claiming exact parity and to state explicitly that an empty value is *deliberately* treated as not satisfying a required secret ("an empty token is no token").

2. **Pattern-failure message (start.go)** — The report now carries `validatePattern`'s actual error instead of hardcoding "does not match required pattern", and the header reads "failed pattern validation". So a bad/unsupported regex reports correctly rather than misleadingly. (In practice meta-load already rejects an uncompilable pattern, but the message is now accurate either way.)

3. **Env leak in `TestSecretFromEnv`** — Added a `t.Cleanup` that saves/restores `SKAINET_TOKEN` around the `os.Unsetenv` subtest, so no env state leaks into later tests in the package.

`go build`, `go vet`, and the full `internal/cli` suite pass.

nhuelstng:
Note on CI: the failing **Test (macos-latest)** check is a pre-existing flaky test, `TestPromptCoalescing` in `internal/netproxy` — a package this PR doesn't touch. It's a race in the test's synchronization (reproduces ~10% under `-race` locally, independent of this branch), not a regression here. Fixed separately in #18; once that lands on main, a re-run here goes green.

nhuelstng:
Rebased onto #18 (the netproxy flake fix) and retargeted this PR's base to that branch, so CI now runs with the fix and the diff here stays limited to the secrets change. **Merge #18 first**; GitHub will then auto-retarget this PR to `main`.

PR #18, opened 2026-06-25
test: de-flake macOS CI (TestPromptCoalescing race + facade health deadline)

nhuelstng:
## Problem

`TestPromptCoalescing` (`internal/netproxy`) is flaky and intermittently fails CI — most visibly on the macOS runner — with:

```
filter_test.go:246: prompter called 3 times; want 1 (coalesced)
```

It reproduces locally in ~10% of runs under `-race`. It is unrelated to whatever PR happens to trigger it; it blocks any PR whose CI lands on an unlucky schedule.

## Root cause

The **production coalescing logic is correct**: a request that arrives while a prompt for the same host is in flight finds the `inflight` entry and waits on `w.done` (`filter.go`). The bug is in the **test's synchronization**.

The test launches 5 concurrent `Check`s for one host, then releases the leader's blocked prompt as soon as `started >= 1`:

```go
waitUntil(t, func() bool { return p.started.Load() >= 1 })
close(block)
```

But `started >= 1` only proves the *leader* reached `Prompt` — it says nothing about the other 4 goroutines. If a straggler reaches `promptCoalesced` *after* the leader's prompt returns and clears `inflight`, it registers a fresh prompt → `started == 2/3` → failure.

## Fix

The test needs to release the leader only once **all followers have parked** on the shared prompt. There's no way to observe that from outside without a seam, so:

- Add an unexported `onCoalesceWait func()` to `FilterConfig` — a test seam, **never set in production** — fired just before a request blocks on an in-flight prompt (the coalescing branch).
- The test counts followers via that hook and waits for all `n-1` to park before `close(block)`. Once a follower has called the hook, it is committed to `<-w.done` (the leader hasn't closed it yet), so it cannot start its own prompt. Deterministic.

No change to production behavior.

## Verification

- `go test -race -count=200 -run TestPromptCoalescing ./internal/netproxy/` → pass (was ~10% fail before).
- `go test -race -count=1 -timeout=5m ./...` (mirrors CI) → all pass.

### DIFF
--- a/internal/facade/integration_test.go
+++ b/internal/facade/integration_test.go
@@ -150,9 +150,12 @@ func TestEchoRestEndToEnd(t *testing.T) {
 		_ = cmd.Wait()
 	})
 
-	// Wait for /status to answer 2xx.
+	// Wait for /status to answer 2xx. The deadline is generous because a
+	// cold Python interpreter start on a loaded CI runner (notably macOS)
+	// can take well over 5s; a tight bound here flaked as "sidecar never
+	// became healthy" even though the sidecar was merely slow to boot.
 	sidecarURL := fmt.Sprintf("http://127.0.0.1:%d/status", sidecarPort)
-	deadline := time.Now().Add(5 * time.Second)
+	deadline := time.Now().Add(30 * time.Second)
 	for time.Now().Before(deadline) {
 		resp, err := http.Get(sidecarURL)
 		if err == nil {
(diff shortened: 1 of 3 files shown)


nhuelstng:
Added a second commit: `TestEchoRestEndToEnd` (internal/facade) gave the Python sidecar only 5s to become healthy, which flaked on loaded macOS runners ("sidecar never became healthy"). Bumped that deadline to 30s — fast runs are unaffected, slow cold-starts no longer false-fail. Both macOS flakes blocking the stack (#15, #17) are now addressed here.

PR #19, opened 2026-06-25
feat(sandbox): add --allow-unix-dir for dynamic AF_UNIX sockets

NoRiceToday:
## Motivation

Claude Code's **Agent View** talks to a local daemon over AF_UNIX sockets under `/tmp/cc-daemon-<uid>/…`, whose leaf names are minted dynamically at runtime (`control.sock`, plus per-session `pty/<worker>.sock`). Under the sandbox's `filtered`/`blocked` network mode, Seatbelt classifies AF_UNIX `connect(2)` as `network-outbound` and denies it — and a single literal `--allow-file` cannot pre-grant a socket whose name does not exist yet. The result: Agent View is unusable inside the omac sandbox.

## Change

Add `--allow-unix-dir <dir>` / `filesystem.allow_unix_dir`, granting AF_UNIX connect to **any** socket under a directory (a path-scoped `subpath` rule) plus read+write access to the dir. Unlike `allow`/`read`/`write`, the path is **not** existence-filtered — the daemon may create the dir after launch.

## Implementation Decisions

- **Path-scoped, never blanket.** The rule is a `subpath` of the named dir, so it cannot reach other host sockets (`docker.sock`, `ssh-agent`, the omac facade socket itself); a path filter matches only AF_UNIX and opens no TCP/IP egress. A guard test forbids the over-broad forms (`(allow network-outbound)`, `subpath "/"`, `network*`).
- **Linux graceful degradation.** AF_UNIX is not Landlock-filtered, so the dir only needs a filesystem bind. Because it may be absent at launch, it is bound with `--bind-try` — a plain `--bind` of a missing source aborts the whole bwrap launch. A missing `/tmp` dir also no longer suppresses the `/tmp` tmpfs fallback.

## Changes Overview

- `flags.go` / `sandbox_cmd.go` — new `--allow-unix-dir` flag (parse, additive merge, help text).
- `profile.go` — `filesystem.allow_unix_dir` field.
- `grants.go` — resolve into `Grants.UnixSocketDirs` (expanded, *not* existence-filtered) plus file access.
- `sbpl.go` (macOS) — emit `(allow network-outbound (subpath …))`; ancestor-canonicalize nonexistent paths.
- `bwrap.go` (Linux) — `--bind-try` for unix dirs; preserve the `/tmp` tmpfs fallback.
- Docs — README grant table, openspec spec (field, flag enumeration, scenario).

## Diagram

```mermaid
flowchart LR
  A["--allow-unix-dir<br/>allow_unix_dir"] --> B["Profile.Filesystem.AllowUnixDir"]
  B --> C["ResolveGrants<br/>Grants.UnixSocketDirs<br/>(+ AllowPaths, not existence-filtered)"]
  C --> D["macOS: GenerateSBPL<br/>(subpath) network-outbound<br/>+ ancestor canonicalization"]
  C --> E["Linux: BuildBwrapArgv<br/>--bind-try dir<br/>(AF_UNIX not Landlock-filtered)"]
```

## Deployment / Impact

- Purely additive and opt-in — a new flag/field; no default profile enables it. No breaking changes to the profile schema or CLI.
- No new egress: the grant is destination-scoped to the named dir; the top-level `(deny network*)` (macOS) and Landlock TCP rules (Linux) are untouched.

## Testing

- New unit tests in `internal/sandboxprofile` and `internal/sandboxrun` cover: flag parse + additive merge; resolution including the nonexistent-dir case; the SBPL `subpath` emission; the path-scoped guard (no over-broad rules); macOS ancestor-canonicalization (portable, symlink-based so it runs on both runners); and the Linux `--bind-try` + tmpfs behavior.
- `gofmt`, `go vet`, `go build`, `staticcheck`, and `go test -race ./...` all clean locally (Go 1.25).
- Validated end-to-end on macOS: with `allow_unix_dir: ["/tmp/cc-daemon-<uid>"]` in a `filtered` profile, Claude Code's Agent View connects to the daemon and its per-session worker sockets inside the sandbox — previously impossible.


nhuelstng:
@copilot resolve the merge conflicts in this pull request

PR #38, opened 2026-07-05
37 add security audit trail

thomas-endres-tng:
omac is a security boundary but kept no unified, tamper-evident record of what the confined agent actually did. This adds a central, append-only, structured (JSON Lines) audit trail covering every security-relevant action omac performs:

Commands — the sandboxed inner command and every spawned sidecar (process.exec / process.exit), with argv (secrets redacted), cwd, exit code, and duration.
Network — every outbound allow/deny decision with its source (net.decision: prompt / learned / allowlist / blocklist / hard-deny / timeout).
Facade requests — method, mount, hashed namespace, upstream status, bytes, duration (facade.request, supersedes the ad-hoc access log).
Control-plane mutations — activate / deactivate / reload / reload-global (control.mutation).
Secrets & routes — which skill received which secret names, never values (secret.inject); non-ready route transitions (route.state).
Session lifecycle — session.start / session.stop.
Key properties

Redaction at a single emit boundary: secret values are never written verbatim, and per-directory namespace tokens are hashed (ns_…) — no call site can leak them.
Persistent central location that survives restarts and sits outside the sandbox's writable grants: Linux ~/.local/state/omac/audit/audit.jsonl (XDG_STATE_HOME honored), macOS ~/Library/Logs/omac/audit/audit.jsonl; dir 0700, file 0600. Runs are distinguished by a per-run run_id; a monotonic seq makes gaps detectable.
Fail-open by default (a write error degrades to one stderr warning); --audit-strict makes an unwritable log fatal (fail-closed) for compliance/forensics postures. --no-audit + --audit-strict is rejected as misuse.
Configurable via the launcher audit block (enabled/path/syslog/strict) and CLI flags (--audit-log, --no-audit, --audit-strict); optional syslog mirror (Unix, build-tagged).
Implementation

New internal/audit/ package (interface + no-op default, event model, redactor, persistent-path resolver, file + optional syslog sinks). Emit points wired into supervisor, the sandbox launcher, the netproxy filter (threaded into the re-exec'd omac sandbox run subprocess so net decisions land in the same file), facade, and the serve control plane. Designed and specced via OpenSpec (openspec/changes/add-audit-trail, now archived; baseline spec at openspec/specs/audit-trail/spec.md).

Testing

Unit tests: envelope/seq monotonicity, redaction (argv secret values + namespace tokens), 0600/0700 permissions, append-across-restart, fail-open-warns-once, strict-mode fatal handler, valid-JSON-per-line, config parsing.
CLI integration test drives control-plane mutations and asserts events + hashed namespace + no token leak; --no-audit --audit-strict misuse.
Verified live: a real omac serve run wrote session.start to the correct persistent path with 0600/0700; a manual emit of all event types confirmed redaction and monotonic sequencing.
go build ./... and go vet ./... clean. Remaining netproxy/sandboxrun/doctor test failures are pre-existing and environmental (loopback TCP / real-sandbox exec / harness binaries unavailable in the sandbox) — confirmed present on the base commit without this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,55 @@ facade:
   idle_timeout_secs: 300
   max_body_bytes: 10485760
   base_env_passthrough: [PATH, HOME, USER, LANG, LC_ALL, LC_CTYPE, TMPDIR]
+audit:
+  enabled: true                     # security audit trail (default on)
+  path: ""                          # "" => persistent central path (see below)
+  syslog: false                     # also mirror events to the system log (Unix)
+  strict: false                     # fail-closed: abort if the log can't be written
 ```
+
+#### Audit trail
+
+omac records a security audit trail: an append-only, structured
+(JSON Lines) log of every security-relevant action it performs — the
+sandboxed inner command and each sidecar it spawns (`process.exec` /
+`process.exit`), outbound network allow/deny decisions and their source
+(`net.decision`), facade requests (`facade.request`), control-plane
+mutations (`control.mutation`), secret injection by name
+(`secret.inject`), non-ready routes (`route.state`), and session
+lifecycle (`session.start` / `session.stop`).
+
+The log lives at a **persistent, central** location that survives
+restarts (successive runs append, distinguished by a per-run `run_id`):
+
+| Platform | Default path |
+|---|---|
+| Linux | `$XDG_STATE_HOME/omac/audit/audit.jsonl` → `~/.local/state/omac/audit/audit.jsonl` |
+| macOS | `~/Library/Logs/omac/audit/audit.jsonl` |
+
+The directory is `0700` and the file `0600`, and the default location is
+**outside** the sandbox's writable grants, so the confined process cannot
+tamper with the host's audit trail. Secret values and per-directory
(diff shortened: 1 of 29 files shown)


nhuelstng:
Review findings fixed in #43:

- **run_id correlation**: subprocess inherited fresh run_id, breaking per-run correlation. Now passes parent run_id + mode via `--audit-run-id`/`--audit-mode`. Seq stays per-process (pid distinguishes writers).
- **Prompt scope/persisted lost**: `Verdict` had no scope/persisted fields; prompt decisions logged `scope=""`, `persisted=false` always. Now propagated from `PromptResult`.
- **process.exit only on teardown**: self-terminated sidecars had no event until `ShutdownAll`. Added `watchChild` reaper; `auditExit` idempotent via `exitOnce` and centralizes `LogFile.Close`.
- **Non-spec sources**: `hard-deny`/`dns`/`default` added to spec enum + spec doc.
- **Mode hardcoded**: sandboxrun hardcoded `ModeStart`; now inherits parent's mode.

nhuelstng (approved):

PR #44, opened 2026-07-06
test(sandbox): add worktree edge-case and Linux integration tests

nhuelstng:
Supplements PR #31 with tests for edge cases and the Linux backend identified during review.

> **Depends on #31.** These tests exercise code added in #31 (`gitWorktreeGrants`, `resolveWorktreeCommonDir`, `pathWithinRoot`). Merge #31 first, then rebase this PR onto main. The tests will fail to compile without #31's changes.

## What

Adds 7 tests covering gaps found in the PR #31 review:

### Unit tests (`grants_test.go`)

| Test | Covers |
|------|--------|
| `TestResolveGrantsSubmoduleNoWorktreeGrants` | Submodule admin dir has no `commondir` → no worktree grants leak into parent common dir. Pins the PR body's submodule exclusion claim. |
| `TestResolveGrantsWorktreeAccessWrite` | `AccessWrite` (write-only) broadens worktree subdirs to read+write (git must read to commit); config/info/packed-refs/hooks stay read-only. |
| `TestResolveGrantsBareRepoWorktree` | Worktree from a bare repo resolves through the git-invariant check (common dir is the bare repo dir itself). |
| `TestResolveGrantsConcurrentWorktreesIsolation` | Two worktrees sharing one common dir each get grants for their OWN admin dir only — never the sibling's. |

### Linux integration tests (`integration_worktree_linux_test.go`, new file)

| Test | Covers |
|------|--------|
| `TestIntegrationWorktreeHooksRunButNotWritable` | Linux bwrap counterpart of the darwin test: sandboxed commit succeeds, prepare-commit-msg hook runs, hooks dir not writable. |
| `TestIntegrationWorktreeSymlinkEscape` | Planted `objects -> secret` symlink blocked under real bwrap. |
| `TestIntegrationWorktreeKnownLimitations` | PR body's known-limitation claims: `branch -d` fails cleanly (ref preserved), `gc` no-ops, commit succeeds despite `packed-refs.lock` EPERM, `fsck` clean. |
| `TestIntegrationWorktreeLandlockCombined` | Combined bwrap + Landlock path: worktree commit succeeds (fs binds) AND network stays blocked (Landlock holds) in the same sandbox. |

## Why

The PR shipped a darwin-only integration test and unit tests covering `AccessReadWrite`/`AccessRead`/`AccessNone`. The review found:

1. **No Linux integration test** — the bwrap backend (bind mounts) is a different code path from SBPL; the unit test checks the argv but doesn't run bwrap.
2. **Submodule exclusion untested** — claimed in the PR body but not pinned by a test.
3. **`AccessWrite` untested** — the worktree path broadens it to read+write, but no test covered this.
4. **Bare-repo worktrees untested** — common CI/git-server layout.
5. **Concurrent worktrees untested** — sibling admin dir isolation not pinned.
6. **Landlock + worktree combined path untested** — the two enforcement layers (bwrap binds + Landlock net) had no combined test.

## Verification

All tests pass on Linux (bwrap 0.11.0, git 2.53.0, Landlock ABI 6):

```
go test ./internal/sandboxrun/... -count=1
ok  github.com/tngtech/oh-my-agentic-coder/internal/sandboxrun  2.4s
```

Build and vet clean. No regressions in the existing suite.

## Review findings from #31 still outstanding

Notes for the PR #31 author, **not** fixed here (out of scope for a test-only PR):

- **PR description mismatch**: body says `deny: hooks/` but code makes hooks read-only (readable, not writable, not hard-denied). The code is better than described — update the description.
- **Dead plumbing**: `denyAdds`/`worktreeDeny` in `gitWorktreeGrants` is always `nil`; the third return value and the `protected = append(protected, worktreeDeny...)` line can be removed.

### DIFF
--- a/internal/sandboxrun/grants_test.go
+++ b/internal/sandboxrun/grants_test.go
@@ -3,6 +3,7 @@ package sandboxrun
 import (
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"slices"
 	"strings"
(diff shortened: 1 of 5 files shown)


NoRiceToday:
LGTM, thanks for the additional test coverage!

NoRiceToday (approved):

internal/sandboxrun/integration_worktree_linux_test.go
```diff
@@ -0,0 +1,374 @@
...
+	t.Helper()
+	c := exec.Command("git", args...)
+	c.Dir = dir
+	c.Env = append(os.Environ(),
+		"GIT_CONFIG_GLOBAL=/dev/null", "GIT_CONFIG_SYSTEM=/dev/null",
+		"GIT_AUTHOR_NAME=t", "GIT_AUTHOR_EMAIL=t@t", "GIT_COMMITTER_NAME=t", "GIT_COMMITTER_EMAIL=t@t")
+	if out, err := c.CombinedOutput(); err != nil {
+		t.Fatalf("git %v: %v\n%s", args, err, out)
+	}
+}
```
NoRiceToday: [Wichtig] **`gitInDir` is duplicated verbatim** in `integration_worktree_darwin_test.go:17` and here. The two files are in the same package, so only one definition compiles per build-tag set — but on a platform that builds *both* (none currently, since the tags are `darwin`/`linux`) they'd collide. More importantly, it's ~12 lines duplicated. Consider moving `gitInDir` to a shared `integration_test.go` (no build tag) or a `testhelpers_test.go`.

internal/sandboxrun/grants_test.go
```diff
@@ -191,6 +192,661 @@ func listenUnix(path string) (interface{ Close() error }, error) {
...
+		t.Skip("git not available")
+	}
+	base := t.TempDir()
+	mainRepo := filepath.Join(base, "main")
+	run := func(dir string, args ...string) {
+		t.Helper()
+		cmd := exec.Command(git, args...)
+		cmd.Dir = dir
+		cmd.Env = append(os.Environ(),
+			"GIT_AUTHOR_NAME=t", "GIT_AUTHOR_EMAIL=t@t.t",
```
NoRiceToday: [Wichtig] **The `run` closure is copy-pasted** across `TestResolveGrantsRealGitWorktreePipeline`, `TestResolveGrantsSubmoduleNoWorktreeGrants`, `TestResolveGrantsBareRepoWorktree`, and `TestResolveGrantsConcurrentWorktreesIsolation`. Each defines the same `func(dir string, args ...string)` with identical env setup. Extract a package-level `testGitRun(t, dir, args...)` helper (or reuse `gitInDir` from the integration files by moving it to a shared helper).

internal/sandboxrun/grants_test.go
```diff
@@ -191,6 +192,661 @@ func listenUnix(path string) (interface{ Close() error }, error) {
...
+	common := filepath.Join(mainRepo, ".git")
+	if resolved, rerr := filepath.EvalSymlinks(common); rerr == nil {
+		common = resolved
+	}
+	for _, list := range [][]string{g.ReadPaths, g.WritePaths, g.AllowPaths} {
+		for _, gp := range list {
+			if gp == common || strings.HasPrefix(gp, common+string(filepath.Separator)) {
+				if !strings.HasPrefix(gp, subPath) {
+					t.Errorf("submodule leaked a grant into the parent common dir: %s", gp)
+				}
```
NoRiceToday: [Hinweis] **Assertion is slightly weaker than the claim.** The comment says "No grant may reach the parent repo's common dir via the submodule", but the condition `if !strings.HasPrefix(gp, subPath)` *excludes* grants under `subPath` from the check. Since `subPath` = `mainRepo/sub` and `common` = `mainRepo/.git`, a grant under `subPath` can never be under `common` anyway — so the `!HasPrefix(gp, subPath)` guard is effectively dead. The check still works (the `HasPrefix(gp, common+sep)` outer guard is what catches leaks), but the inner guard suggests an intent that doesn't apply. Consider simplifying to just the `common` prefix check.

internal/sandboxrun/integration_worktree_linux_test.go
```diff
@@ -0,0 +1,374 @@
...
+	// 1. Commit must succeed even with packed-refs present (non-fatal EPERM).
+	out, code := sandboxGit("commit", "--allow-empty", "-m", "c1")
+	if code != 0 {
+		t.Fatalf("commit failed (exit %d) — packed-refs.lock EPERM must be non-fatal:\n%s", code, out)
+	}
+	if strings.Contains(strings.ToLower(out), "fatal") {
+		t.Errorf("commit produced fatal output:\n%s", out)
+	}
+
+	// 2. git gc should no-op or succeed, not corrupt.
```
NoRiceToday, line 230: [Hinweis] **`gc` assertion is lenient.** `if code != 0 && strings.Contains(..., "fatal")` only errors when *both* non-zero *and* contains "fatal". A non-zero exit without "fatal" in output is silently accepted (intentional — gc may warn), and a zero-exit with "fatal" in output (unlikely) is also accepted. A one-line comment like `// gc may warn (non-fatal); only flag fatal failures` would document the intent.

internal/sandboxrun/integration_worktree_linux_test.go
```diff
@@ -0,0 +1,374 @@
...
+	repo := filepath.Join(base, "repo")
+	if err := os.MkdirAll(repo, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	gitInDir(t, repo, "init", "-q")
+	if err := os.WriteFile(filepath.Join(repo, "seed"), []byte("x\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	gitInDir(t, repo, "add", "seed")
+	gitInDir(t, repo, "commit", "-q", "-m", "seed")
```
NoRiceToday, line 289: [Hinweis] **Widening read surface.** `g.ReadPaths = append(g.ReadPaths, filepath.Dir(omac))` grants read on the whole temp dir holding the `omac` binary, not just the binary. `BuildBwrapArgv` will `--ro-bind` the entire dir. Narrow to the binary: `g.ReadPaths = append(g.ReadPaths, omac)`. (The binary itself is sufficient for `stage2` to exec it.)

internal/sandboxrun/integration_worktree_linux_test.go
```diff
@@ -0,0 +1,374 @@
...
+	wt := filepath.Join(base, "wt")
+	gitInDir(t, repo, "worktree", "add", "-q", wt)
+
+	secret := filepath.Join(base, "secret")
+	if err := os.MkdirAll(secret, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	target := filepath.Join(secret, "pwned")
+
+	// Plant the symlink: objects -> secret. The real objects dir is moved
```
NoRiceToday, line 137: [Hinweis] **`objects/` moved aside but not restored.** `os.Rename(realObjects, filepath.Join(base, "real-objects"))` leaves the repo structurally invalid (git would refuse to operate). `t.Cleanup` removes `base`, so no leak, but a one-line note that the repo is deliberately broken after this point would help future readers.

internal/sandboxrun/grants_test.go
```diff
@@ -191,6 +192,661 @@ func listenUnix(path string) (interface{ Close() error }, error) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	admin := filepath.Join(common, "worktrees", "wt")
+	wantAllow := []string{
+		filepath.Join(common, "objects"),
+		filepath.Join(common, "refs"),
+		admin,
+	}
+	// logs/ exists only when core.logAllRefUpdates is set (default off
```
NoRiceToday, line 729: [Hinweis] **Good defensive coding.** Conditionally checking `logs/` only when it exists (bare repos default `core.logAllRefUpdates=off`) is the right call, and the comment explains the why. No change needed — flagging as a positive example.

nhuelstng:
adds on top of #31 and closes it

PR #46, opened 2026-07-07
feat(sandbox): protect workdir .env/.envrc by default

nhuelstng:
## What

Workspace-local `.env` and `.envrc` files are now denied by default inside the sandbox, without needing `--deny .env` or `filesystem.deny`.

## Why

The baseline protected-path set covered `~/.env` (home dir), but a workdir-local `.env` holding real secrets was readable by the sandboxed agent unless the user remembered to opt in via `--deny .env`. This is a common footgun — workspace `.env` is the most likely place for project secrets.

## How

- New `WorkdirProtected []string` field on `Baseline` (`internal/sandboxprofile/baseline.go`) — basenames resolved at launch time against the workdir + every explicitly granted tree.
- `resolveBaselineWorkdirDeny` in `internal/sandboxrun/grants.go` reuses the existing `walkGlobMatches` machinery (same as `filesystem.deny` globs), so baseline system trees (`/usr`, `/lib`) are never scanned — only the workdir and explicit user grants.
- `filesystem.override_deny` holes punch through in both basename (`.env`) and absolute-path forms, so a skill that legitimately needs to read a workdir `.env` can opt out.

## Files

- `internal/sandboxprofile/baseline.go` — `WorkdirProtected` field + `workdirProtectedCommon()` (.env, .envrc)
- `internal/sandboxrun/grants.go` — `resolveBaselineWorkdirDeny` + wiring in `ResolveGrants`
- `internal/sandboxrun/grants_test.go` — `TestResolveGrantsWorkdirEnvProtectedByDefault` (default protection + override_deny basename + override_deny absolute)
- `README.md` — table row for workdir `.env`/`.envrc`

## Verification

- `go test ./internal/sandboxrun/... ./internal/sandboxprofile/...` — pass
- `go vet ./...` — clean
- gofmt — clean
- Existing deny tests still pass (no regression in `filesystem.deny` / `--deny` behavior)

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ sandbox:
 | Bridge socket (`$TMPDIR/omac-<hash>/bridge.sock`) | read+write | `--allow-file` / `--read` flags |
 | Dynamic socket dir (e.g. Agent View `/tmp/cc-daemon-<uid>`) | read+write + AF_UNIX connect | `--allow-unix-dir` flag / `filesystem.allow_unix_dir` |
 | Paths in `~/.ssh`, `~/.gnupg`, `~/.aws`, `~/.kube`, … | **denied** | protected paths (override with `filesystem.override_deny`) |
-| Workdir `.env` and `.envrc` (incl. nested) | **denied** | baseline workdir-protected set (override with `filesystem.override_deny: [".env"]`) |
+| Workdir and granted-tree `.env` / `.envrc` (incl. nested) | **denied** | baseline workdir-protected set (override with `filesystem.override_deny: [".env"]`) |
 | Files matching `filesystem.deny` (e.g. `*.key`) inside granted trees | **denied** | user deny list (`filesystem.deny` / `--deny`) |
 
 ## Typical workflow
--- a/README.md
+++ b/README.md
@@ -402,7 +402,8 @@ sandbox:
 | Bridge socket (`$TMPDIR/omac-<hash>/bridge.sock`) | read+write | `--allow-file` / `--read` flags |
 | Dynamic socket dir (e.g. Agent View `/tmp/cc-daemon-<uid>`) | read+write + AF_UNIX connect | `--allow-unix-dir` flag / `filesystem.allow_unix_dir` |
 | Paths in `~/.ssh`, `~/.gnupg`, `~/.aws`, `~/.kube`, … | **denied** | protected paths (override with `filesystem.override_deny`) |
-| Files matching `filesystem.deny` (e.g. `.env`, `*.key`) inside granted trees | **denied** | user deny list (`filesystem.deny` / `--deny`) |
+| Workdir `.env` and `.envrc` (incl. nested) | **denied** | baseline workdir-protected set (override with `filesystem.override_deny: [".env"]`) |
+| Files matching `filesystem.deny` (e.g. `*.key`) inside granted trees | **denied** | user deny list (`filesystem.deny` / `--deny`) |
 
 ## Typical workflow
 
--- a/internal/sandboxprofile/baseline.go
+++ b/internal/sandboxprofile/baseline.go
@@ -17,6 +17,12 @@ type Baseline struct {
 	// ProtectedPaths are denied even when covered by a broader grant,
 	// unless listed in filesystem.override_deny.
 	ProtectedPaths []string
+	// WorkdirProtected are basenames (e.g. ".env") denied inside the
+	// workdir and any explicitly granted tree. Unlike ProtectedPaths,
+	// they are resolved at launch time against the actual workdir, so
+	// they catch workdir-relative files the static ProtectedPaths set
+	// cannot. override_deny holes punch through here too.
+	WorkdirProtected []string
 }
 
 // PlatformBaseline returns the baseline for the current GOOS.
@@ -72,6 +78,17 @@ func protectedCommon() []string {
 	}
 }
 
+// workdirProtectedCommon are basenames denied inside the workdir and
+// any explicitly granted tree. Resolved at launch time against the
+// actual workdir (see ResolveGrants), so workdir-relative files like
+// `<workdir>/.env` are blocked by default without a --deny flag.
+func workdirProtectedCommon() []string {
+	return []string{
+		".env",
+		".envrc",
+	}
+}
+
 func darwinBaseline() Baseline {
 	return Baseline{
 		Read: []string{
@@ -122,6 +139,7 @@ func darwinBaseline() Baseline {
 			"~/Library/Containers/com.apple.Safari",
 			"~/Library/Application Support/MobileSync",
 		),
+		WorkdirProtected: workdirProtectedCommon(),
 	}
 }
 
@@ -151,6 +169,7 @@ func linuxBaseline() Baseline {
 			"~/.config/microsoft-edge",
 			"~/.config/BraveSoftware",
 		),
+		WorkdirProtected: workdirProtectedCommon(),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


internal/sandboxprofile/baseline.go
```diff
@@ -17,6 +17,12 @@ type Baseline struct {
 	// ProtectedPaths are denied even when covered by a broader grant,
 	// unless listed in filesystem.override_deny.
 	ProtectedPaths []string
+	// WorkdirProtected are basenames (e.g. ".env") denied inside the
+	// workdir and any explicitly granted tree. Unlike ProtectedPaths,
+	// they are resolved at launch time against the actual workdir, so
+	// they catch workdir-relative files the static ProtectedPaths set
+	// cannot. override_deny holes punch through here too.
```
anotherthomas: the "override_deny" comment is here and in line 121. redundant?

anotherthomas: and in a bunch of other places, including the "e.g. .env"

internal/sandboxrun/grants.go
```diff
@@ -179,6 +185,45 @@ func resolveUserDeny(deny, scanRoots []string, notices io.Writer) []string {
 	return out
 }
 
+// resolveBaselineWorkdirDeny resolves the baseline workdir-protected
+// basenames (e.g. ".env") against the granted scan roots, mirroring
+// resolveUserDeny's glob walk. override_deny entries that are bare
```
anotherthomas: could that be the same function with different parameters? "mirroring ..." isn't obvious with a different error handling and significantly different code flow

NoRiceToday:
## Should we also protect `auth.json` in the home data dir?

Came up while thinking about what else should be denied by default.

`auth.json` (the opencode model-provider credential file) is written to `~/.local/share/opencode/auth.json` — see `internal/e2e/harnesses.go:158-172`. That's a **home data dir** path, not a workdir path, so adding it to `WorkdirProtected` would be a no-op: the workdir matcher only walks `dedupe(denyScan)` (workdir + explicit grants) and never reaches `~/.local/share/opencode` unless it's explicitly granted (which it isn't by default).

The real gap: **`~/.local/share/opencode` is not in `protectedCommon()`** — only `~/.local/share/keyrings` is (`baseline.go:163`). So a grant that happened to cover the home dir (or a `--read ~/.local/share/opencode`) would expose `auth.json`, which holds the model API key.

Two options:
1. Add `~/.local/share/opencode/auth.json` (or the whole `~/.local/share/opencode` dir) to `protectedCommon()` in this PR — small, same-family hardening.
2. File as a separate follow-up to keep #46 scoped to workdir protection.

Any preference? Happy to do either.

internal/sandboxrun/grants.go
```diff
@@ -115,6 +115,9 @@ func ResolveGrants(p *sandboxprofile.Profile, workdir string, notices io.Writer)
 	// trees so the same deny covers the cwd and any explicit grant.
 	protected = append(protected, resolveUserDeny(p.Filesystem.Deny, dedupe(denyScan), notices)...)
 
+	// Baseline workdir-protected basenames resolved against the same scan roots.
+	protected = append(protected, resolveBaselineWorkdirDeny(base.WorkdirProtected, p.Filesystem.OverrideDeny, dedupe(denyScan), notices)...)
```
NoRiceToday: These two calls each invoke `walkGlobMatches` on `dedupe(denyScan)`, so every granted tree is walked twice when both user deny globs and baseline basenames are active. `maxDenyScanEntries` bounds the worst case, but it is 2× I/O for large `--allow` trees.

Hoisting `scanRoots := dedupe(denyScan)` once and passing it to both calls fixes the dedupe duplication; merging the globs into a single `walkGlobMatches` call (with per-match attribution) would also halve the I/O.

internal/sandboxrun/grants.go
```diff
@@ -179,6 +182,45 @@ func resolveUserDeny(deny, scanRoots []string, notices io.Writer) []string {
 	return out
 }
 
+// resolveBaselineWorkdirDeny resolves the baseline workdir-protected
+// basenames (e.g. ".env") against the granted scan roots, sharing the
+// walkGlobMatches walk with resolveUserDeny. override_deny entries that
+// are bare basenames (e.g. ".env") or absolute paths matching a found
+// file punch holes in the result, so a skill that legitimately needs to
+// read a workdir .env can opt out via filesystem.override_deny.
+func resolveBaselineWorkdirDeny(basenames, overrideDeny, scanRoots []string, notices io.Writer) []string {
```
NoRiceToday: The override-keying policy now lives in two places: `EffectiveProtectedPaths` (baseline.go:179, for static `ProtectedPaths`) and here (for `WorkdirProtected`), each with its own raw+expanded lookup. If override matching gains glob support or trailing-slash normalization later, one will silently drift.

Worth extracting a shared `buildOverrideLookup(overrideDeny) map[string]bool` helper that both call.

internal/sandboxrun/grants_test.go
```diff
@@ -388,3 +388,84 @@ func TestResolveGrantsDenyGlobDoesNotScanBaseline(t *testing.T) {
 		}
 	}
 }
+
+// TestResolveGrantsWorkdirEnvProtectedByDefault verifies that a
+// workdir-local .env / .envrc is masked by the baseline
+// WorkdirProtected set without any --deny flag, and that
+// filesystem.override_deny can punch a hole through it.
+func TestResolveGrantsWorkdirEnvProtectedByDefault(t *testing.T) {
```
NoRiceToday, line 396: `TestResolveGrantsDenyGlobDoesNotScanBaseline` guards the "baseline system trees never scanned" invariant for *user* deny globs, but there is no parallel test for the *baseline* `.env`/`.envrc` walk. The invariant holds today (shares `denyScan`), but a regression that accidentally passed `base.Read` into the baseline resolver would go undetected.

Would be worth adding a subtest mirroring the user-deny one — assert no protected path starts with `/usr/` or `/lib/`.

README.md
```diff
@@ -402,7 +402,8 @@ sandbox:
 | Bridge socket (`$TMPDIR/omac-<hash>/bridge.sock`) | read+write | `--allow-file` / `--read` flags |
 | Dynamic socket dir (e.g. Agent View `/tmp/cc-daemon-<uid>`) | read+write + AF_UNIX connect | `--allow-unix-dir` flag / `filesystem.allow_unix_dir` |
 | Paths in `~/.ssh`, `~/.gnupg`, `~/.aws`, `~/.kube`, … | **denied** | protected paths (override with `filesystem.override_deny`) |
-| Files matching `filesystem.deny` (e.g. `.env`, `*.key`) inside granted trees | **denied** | user deny list (`filesystem.deny` / `--deny`) |
+| Workdir `.env` and `.envrc` (incl. nested) | **denied** | baseline workdir-protected set (override with `filesystem.override_deny: [".env"]`) |
```
NoRiceToday: The row says "Workdir `.env` and `.envrc`" but the code (`grants.go:119`) also denies them inside every explicitly granted tree (`--read`/`--write`/`--allow`), not just the workdir. The field doc (`baseline.go:20-24`) and `workdirProtectedCommon()` doc both correctly say "workdir and any explicitly granted tree" — the README row is narrower than its own docs.

Suggest: `Workdir and granted-tree .env / .envrc (incl. nested)`.

nhuelstng:
wouldnt blocking the access to auth.json break opencode?

NoRiceToday:
It depends on your setup/how you configured opencodes secrets, I think? But yeah it's probably out of scope here and should be evaluated and tested  in a dedicated issue/mr

NoRiceToday (approved):
LGTM now

Issue #57, opened 2026-07-11
omac register blocks registering a skill under a second harness — pre-check matches by name only, ignoring harness scope


GevatterGaul:
## Summary

  Registering an already-registered skill under a *second* harness fails with
  `already registered with a different bundle_hash; pass --force to update`, even
  though the registry is explicitly designed to hold the same skill under multiple
  harnesses. The pre-check compares the incoming skill against a **different
  harness's** entry, producing a false conflict.

  ## Severity

  Low/medium — a `--force` workaround exists and is non-destructive, but the error
  is misleading and blocks the documented multi-harness workflow.

  ## Environment

  - `omac` (main), Linux
  - A skill (`skill-marketplace`) installed as two separate on-disk copies:
    - `~/.config/opencode/skills/skill-marketplace` — registered under `opencode`
    - `~/.claude/skills/skill-marketplace` — not yet registered

  ## Steps to reproduce

  1. Have `skill-marketplace` registered under the `opencode` harness (entry in
     `~/.config/omac/sidecar.json`).
  2. Have a separate copy of the same skill under `~/.claude/skills/skill-marketplace`
     (different files → different bundle hash).
  3. Run:
     omac register skill-marketplace --harness claude-code

  ## Expected

  The skill registers under the `claude-code` harness as a new, coexisting entry.
  The registry data model supports this: `Registry.Upsert` keys by `(Name, Harness)`
  and `TestUpsertPerHarnessCoexist` proves two harness entries with the same name
  coexist. No `--force` should be required, because no `claude-code` entry exists yet.

  ## Actual

  omac register: registry: already registered with a different bundle_hash; pass --force to update

  ## Root cause

  The pre-check in `internal/cli/register.go:276` uses the **name-only** lookup
  `reg.Find(skillName)` instead of the harness-scoped `reg.FindForHarness(skillName, regHarness)`:

  ```go
  if existing, _ := reg.Find(skillName); existing != nil {
      if existing.BundleHash != bundleHash && !*force {
          return fmt.Errorf("already registered with a different bundle_hash; pass --force to update")
      }
  }

  reg.Find returns the first entry matching the name (the opencode entry here),
  whose bundle_hash differs from the claude-code copy on disk, so the guard
  fires. Registry.Find's own doc comment (internal/registry/registry.go:205-208)
  notes it is back-compat only and that callers with a harness context should
  prefer FindForHarness.

  The write path below the guard is already correct — reg.Upsert (register.go:294)
[...]

PR #58, opened 2026-07-11
fix(register): scope bundle_hash conflict check to the target harness…

GevatterGaul:
… (#57)

`omac register <skill> --harness <h>` refused to register a skill under a second harness when it was already registered under a different one, failing with "already registered with a different bundle_hash; pass --force to update".

The pre-registration guard used the name-only `reg.Find`, which returned the other harness's entry and compared its (naturally different) bundle hash against the incoming copy. The registry already keys entries by (Name, Harness) via `reg.Upsert`, so the two registrations were never in real conflict.

Use `reg.FindForHarness(skillName, regHarness)` so the guard is scoped to the harness being registered. A same-harness re-register with a changed bundle is still refused without --force.

Adds regression tests covering both the new-harness (no --force) and same-harness-changed-bundle (still guarded) paths.

### DIFF
--- a/internal/cli/register.go
+++ b/internal/cli/register.go
@@ -273,7 +273,13 @@ func runRegister(args []string, env *Env) int {
 		if err != nil {
 			return err
 		}
-		if existing, _ := reg.Find(skillName); existing != nil {
+		// Scope the conflict check to the harness we're registering under.
+		// The registry keys entries by (Name, Harness) and the same skill
+		// may legitimately be registered under multiple harnesses (each with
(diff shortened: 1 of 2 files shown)


anotherthomas:
How about the auto-register in serve.go? Does that also need the harness-deambiguation?

nhuelstng:
documented that in #69 
but I think it is fine for merge for now

NoRiceToday:
necro comment: I really don't like redundant comments

PR #65, opened 2026-07-13
fix(keychain): actionable WSL hint on Secret Service failure, doc callout

### LINKED ISSUE #63: WSL: keychain registration fails with unhelpful D-Bus error, no guidance in docs
## What

On WSL, `omac register` fails with:

```
keychain: keychain set omac/<skill>/<SECRET_NAME>: The name org.freedesktop.secrets was not provided by any .service files
```

There is no actionable guidance in the docs or in the CLI output about why this happens or how to fix it.

## Why

`internal/keychain/keychain.go` only supports the OS keychain via D-Bus Secret Service (Linux), macOS Keychain, and Windows Credential Manager (via `go-keyring`). The package doc comment states:

> A file-based fallback for headless Linux is declared as future work in the design doc and is not implemented in v0.

`oh-my-agentic-coder.md` (design doc, §16.2) already describes the intended behavior for this case:

> WSL / headless Linux: Pass-through to the native backend if present; otherwise fallback to an age-encrypted file under `~/.local/share/omac/secrets.age` with the passphrase unlocked once per session via keyring if available.

and mentions `OMAC_KEYRING_BACKEND=secret-service|keychain|file` as a pin. None of this is implemented — `OMAC_KEYRING_BACKEND` and `secrets.age` don't appear anywhere in the codebase (only in the design doc).

WSL typically has no Secret Service provider (gnome-keyring/kwallet) running or installed at all, since it lacks a full desktop session, so this is a very likely first-run failure for any WSL user following the install instructions (`install/install.linux.sh` is symlinked for `wsl` too).

## Ask

1. **Docs**: Add an explicit WSL callout in the install docs / README stating that the keychain backend requires a running Secret Service provider (e.g. gnome-keyring) on WSL, since it isn't installed by default, with a minimal setup snippet (install gnome-keyring + dbus-x11, start/unlock via `dbus-launch` + `gnome-keyring-daemon --unlock --components=secrets`).
2. **CLI hint**: When `omac register`/`omac secrets`/`omac doctor` hits `isKeychainUnavailable` on a host detected as WSL (`internal/osinfo`), print an actionable hint (e.g. "No Secret Service provider found. On WSL this usually means gnome-keyring/kwallet isn't installed or running — see docs/... for setup, or set `DBUS_SESSION_BUS_ADDRESS`.") instead of just the raw D-Bus error.
3. Optionally, reconsider/implement the file-based fallback described in the design doc, or explicitly mark that section as not-yet-implemented so the doc doesn't overpromise.

## Environment

- WSL2, `omac register` (skill marketplace registration)

nhuelstng:
## What
- Export `keychain.IsUnavailable` (was internal `isKeychainUnavailable`) and add `keychain.Ping` for a lightweight backend-reachability probe.
- Add `cli.wrapKeychainErr` / `keychainUnavailableHint`: append an OS-aware, actionable hint (WSL gets a gnome-keyring setup snippet) to write-path keychain errors, without touching read-path behavior (`env_passthrough`/optional-secret fallbacks still work).
- Wire the hint into `omac register`, `omac secrets set`, `omac secrets import`, and add an upfront `[ok]/[warn] keychain backend` line to `omac doctor`.
- README: WSL prerequisites callout with the gnome-keyring setup snippet.
- Design doc §16.2: mark the age-encrypted file fallback and `OMAC_KEYRING_BACKEND` as not implemented, so it stops overpromising.

## Why
On WSL/headless Linux without a running Secret Service provider, `omac register`/`omac secrets` surfaced the raw D-Bus error (`org.freedesktop.secrets was not provided by any .service files`) with no guidance, and `omac doctor` silently counted secrets as "missing" instead of explaining why. Fixes #63.

The design doc's file-based fallback (`secrets.age`, `OMAC_KEYRING_BACKEND`) was never implemented, so this PR scopes to docs + CLI hint (issue ask #3's "optional" scope) rather than building that out — it's a meaningfully larger feature (new crypto dependency, passphrase UX, second secret-storage code path) that deserves its own design/issue.

## How
- `keychain.IsUnavailable` classifies backend-level errors (missing Secret Service, no D-Bus session) as opposed to per-secret failures. It was already used on the read path (`GetScoped`/`HasScoped` map it to `ErrNotFound`); this PR reuses it on the write path to attach a hint instead of silently swallowing it.
- `wrapKeychainErr` only touches errors that classify as unavailable — genuine per-secret failures (permission denied, etc.) pass through unchanged.
- `keychain.Ping` performs a lookup for a secret that can never exist; a not-found result means the backend answered, any other error is the backend failure itself.

## Verification
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- Added `TestIsUnavailable` (keychain), `TestWrapKeychainErrOnlyHintsBackendFailures` and `TestKeychainUnavailableHintMentionsWSLSetup` (cli).
- Manually ran `omac doctor` on this host (reports `[ok] keychain backend: reachable`) and with a broken `DBUS_SESSION_BUS_ADDRESS` (confirmed the `[warn]` path fires with the raw error surfaced, i.e. wiring executes correctly end to end).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,54 @@ jobs:
       - name: Test
         run: go test -race -count=1 -timeout=5m ./...
 
+  wsl2-session:
+    # WSL2-style degraded session. WSL2 distros commonly run without systemd
+    # (needs systemd=true in wsl.conf), and even with it, `Linger=no` makes
+    # systemd tear down /run/user/<uid> — including the session-bus socket —
+    # when the login session ends, while DBUS_SESSION_BUS_ADDRESS still points
+    # at the now-dead socket. Real WSL2 isn't worth running on CI (hosted
+    # windows runners only give WSL1; WSL2 needs nested virtualization), but
+    # the failure mode reproduces deterministically here: point the bus address
+    # at a socket that doesn't exist and confirm omac's keychain/Secret Service
+    # reads degrade gracefully (keychain.IsUnavailable) instead of surfacing
+    # raw dial errors.
+    name: WSL2-style session (dead bus)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install bubblewrap
+        run: |
+          sudo apt-get update && sudo apt-get install -y bubblewrap
+          # This job degrades the session, not the sandbox, so bwrap still
+          # needs the unprivileged-userns permission AppArmor denies by default
+          # on Ubuntu 24.04 — otherwise the sandbox integration tests would
+          # skip rather than run under the degraded session.
+          sudo tee /etc/apparmor.d/bwrap > /dev/null <<'EOF'
+          abi <abi/4.0>,
+          /usr/bin/bwrap flags=(unconfined) {
+            userns,
+          }
+          EOF
+          sudo apparmor_parser -r /etc/apparmor.d/bwrap
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Test under a torn-down session bus
+        env:
+          # A /run/user/<uid>/bus that systemd already removed: the address is
+          # still exported, but the socket is gone (Linger=no teardown).
+          DBUS_SESSION_BUS_ADDRESS: unix:path=${{ runner.temp }}/dead-user-bus
+          # The runtime dir was never created — mimics a session with no
+          # systemd --user instance to populate it.
+          XDG_RUNTIME_DIR: ${{ runner.temp }}/nonexistent-run-user
+        run: go test -race -count=1 -timeout=5m ./...
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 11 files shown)


nhuelstng:
## Added commit 0759cdd — dead session-bus classification + WSL2 CI job

**What** — Extends `IsUnavailable` to classify a torn-down session-bus socket (`dial unix …: connect: no such file or directory` / `connection refused` / `permission denied`), plus a real go-keyring end-to-end test and a `wsl2-session` CI job.

**Why** — The classifier only matched `dbus`/`org.freedesktop.secrets` markers. On WSL2 without systemd / `Linger=no`, systemd removes `/run/user/<uid>` while `DBUS_SESSION_BUS_ADDRESS` still names the dead socket; go-keyring then returns a bare dial error with **no** dbus marker. Read paths (`GetScoped`/`HasScoped`) surfaced it as a hard error instead of degrading to `ErrNotFound`, and the new write-path hint never fired for that case. The old `TestIsUnavailable` "no dbus session" case passed only because its hand-written string contained `dbus` — not what go-keyring actually emits.

**How** — Match scoped to `dial unix` so genuine per-secret errors are never masked. CI job points the bus at a non-existent socket + absent `XDG_RUNTIME_DIR`, keeping bubblewrap real.

**Verification** — New tests fail without the fix; full suite green under the dead-bus env; gofmt/vet clean; YAML valid. Closes #94.

Issue #66, opened 2026-07-13
e2e: testing ideas to validate (security + creative)


nhuelstng:
## Context

Brainstorm from a review of the current e2e suite (`internal/e2e/`). Flakiness-reduction ideas are being addressed directly (see PR); this issue tracks the security-testing and other creative ideas for future validation/prioritization — not yet scoped or estimated.

Known gaps already called out in code comments (`e2e_test.go`) as documented-but-not-asserted:
- Cross-skill sidecar isolation — `logCrossSkillIsolation`
- Exec-on-read-only-mount behavior — `logExecProbeResults`
- `--no-sandbox` harnesses (codex/macOS) skip all negative assertions entirely

## Security testing ideas

- [ ] Promote cross-skill sidecar isolation from a logged observation to a hard assertion (or explicitly decide it's out of scope and document why).
- [ ] Promote exec-on-read-only-mount from logged to asserted, or document as an accepted platform default.
- [x] Expand `AllowanceSpec` (`allowance.go`) fs-deny targets: cloud creds (`~/.azure`, `~/.gcloud`), and critically the host Docker socket (`/var/run/docker.sock`) — a container-escape vector not currently covered. Done in #84 (also fixed a pre-existing rigor gap in the fs assertions themselves — see comment below).
- [x] Symlink/hardlink escape test: agent creates a symlink inside the allowed workdir pointing at a denied path, then reads/writes through it. Implemented: `symlink` probe in `audit.sh` + `assertSymlinkEscapeDenied` (hard assertion, read+write through a workdir symlink to a denied path); `hardlink` probe + `logHardlinkProbeResults` (log-only — hardlink creation requires same-filesystem as target, so failure can mean EXDEV rather than a sandbox decision).
- [ ] Network pivot tests: DNS-exfiltration attempt, allowed-domain-wrong-path SSRF-style probe, and an IPv6-route regression test tied to the netproxy pinned-address fix (bd2a5c9) — verify a dead IPv6 route can't bypass proxy pinning.
- [ ] Don't skip negative assertions entirely for `--no-sandbox` harnesses (codex/macOS) — at least assert the audit *reporting* correctly flags exposure, so the real risk surface for those harnesses is documented instead of silently excluded.
- [ ] Provenance tampering test: a malicious skill tries to forge/suppress `omac provenance --json` records; assert the mismatch is caught (builds on `provenance_test.go`).
[...]

PR #100, opened 2026-07-15
test(e2e): run model-free e2e tests at PR time; assert --no-sandbox exposure

nhuelstng:
## What

- Add an `e2e_fast` build tag on the model-free files in `internal/e2e/` (`//go:build e2e || e2e_fast`) and a new **E2E fast** CI job that runs just that slice on `ubuntu-latest` + `macos-latest` on every push/PR.
- Close the `--no-sandbox` gap: `runSecurityAudit` no longer silently skips all negative assertions — it asserts the audit ran to completion and documents the per-property exposure surface.
- Refactor the pure assertion predicates and `buildOmac` into shared, fast-buildable files.

## Why

The entire `internal/e2e/` package sits behind `//go:build e2e`, and that tag is compiled **only** by `e2e.yml` (weekly cron + manual dispatch). `ci.yml` and `release.yml` run `go test ./...` without it, so the model-free security-regression tests never ran on the merge path — even the ones explicitly built to run fast without a live model:

| Test | Guards |
|------|--------|
| `TestE2EServeDirTokenIsolation` | the real #74 dir_token cross-workdir leak, over real loopback sockets |
| `security_assertions_test.go` | assertions silently becoming no-ops (the #84 rigor gap / #66 "mutation-test the tests") |
| `versions_test.go`, `harnesses_test.go` | harness/version registry invariants |

Consequence: a reintroduced dir_token leak or a neutered fs-deny assertion would pass CI, pass the release gate, and ship — surfacing up to a week later. This moves the fast net onto the code path that produces releases.

For `--no-sandbox` harnesses (codex/macOS), the audit previously skipped every negative assertion, silently excluding the real risk surface (#66). It now fails loudly if the audit produced nothing and logs an explicit EXPOSED/contained table per property.

## How

- `e2e || e2e_fast` on `harnesses.go`, `versions.go`, `classify.go`, `fixtures.go` (new), `security_assertions.go` (new), and the four model-free test files. The live files (`e2e_test.go`, `allowance.go`, `artifacts.go`, `provenance_test.go`) stay `e2e`-only.
- Full live suite is unchanged: those files build under both tags, so `e2e.yml`'s `-tags=e2e` still runs everything weekly.
- `noSandboxExposureReport` is a pure function unit-tested in the fast slice (`TestNoSandboxExposureReport`), in the repo's "mutation-test the tests" spirit.
- No bubblewrap / token / harness install in the fast job — `omac serve --no-inner` is control-plane only.

## Verification

```
go test -tags=e2e_fast -race ./internal/e2e/   # 11 tests pass (~3s), incl. real serve subprocess
go vet  -tags=e2e       ./internal/e2e/         # full live suite still compiles
go build ./... && go vet ./...                  # untagged build unaffected
```

## Follow-up

A separate release-procedure gap (release currently gates only on untagged `go test`) will be filed as its own issue.

Refs #66

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,3 +141,31 @@ jobs:
         with:
           version: latest
           install-go: false
+
+  e2e-fast:
+    # Model-free slice of the e2e suite (build tag `e2e_fast`): the security
+    # regression tests that need no live LLM harness, API token, or OS
+    # sandbox — dir_token isolation driven over real loopback sockets, the
+    # mutation-tested assertion predicates, the --no-sandbox exposure report,
+    # and the harness/version registry invariants. The full e2e.yml suite
+    # (live models) still runs weekly; this gate keeps these regressions off
+    # the merge path instead of surfacing them up to a week later. No
+    # bubblewrap: `omac serve --no-inner` exercises the control plane only.
+    name: E2E fast (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Run model-free e2e tests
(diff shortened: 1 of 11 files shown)


nhuelstng:
Follow-up release-procedure gap filed as #101 (release currently gates only on the untagged `go test`; the `e2e_fast` slice added here doesn't yet run before a tag is cut).

PR #104, opened 2026-07-15
feat(isolation): Isolate sandbox tool caches for #26

### LINKED ISSUE #26: Sandbox: isolate tool cache from ~/.cache (security + usability)
## Problem

The default sandbox profile grants `~/.cache` read-write to every harness session. `~/.cache` is shared across all applications and can contain sensitive cached data:

- `~/.cache/gh/` — GitHub CLI OAuth token cache, API response cache
- `~/.cache/copilot/` — Copilot package cache
- `~/.cache/claude-cli-nodejs/` — Claude CLI staging
- Any other app's cached data

A compromised or buggy harness running inside the sandbox can read this data (cross-app leakage) or poison cached files for other apps.

## Proposed approach

Replace the broad `~/.cache` rw grant with a **per-workdir persistent cache dir** at `~/.cache/omac/<workdir-hash>/`. Redirect specific tool caches into it via env vars:

```go
func cacheEnvVars(cacheDir string) map[string]string {
    return map[string]string{
        "GOCACHE":         filepath.Join(cacheDir, "go-build"),
        "NPM_CONFIG_CACHE": filepath.Join(cacheDir, "npm"),
        "PIP_CACHE_DIR":   filepath.Join(cacheDir, "pip"),
        "CARGO_HOME":      filepath.Join(cacheDir, "cargo"),
    }
}
```

`XDG_CACHE_HOME` is **not** set — it would redirect ALL XDG-respecting tools, breaking those that need existing cache data from other apps (e.g. `gh` hardcodes `~/.cache/gh/`).

The cache dir is created by omac at launch, granted `--allow` (rw), and persists across sessions for warm rebuilds. Tools without env var overrides keep their default paths — users add specific cache dirs to their profile if needed.

### Implementation sketch

1. `createCacheDir(workdir)` — creates `~/.cache/omac/<sha256(workdir)[:6]>/` (0o700)
2. Grant `--allow <cacheDir>` in start.go + serve.go
3. Set `GOCACHE`, `NPM_CONFIG_CACHE`, `PIP_CACHE_DIR`, `CARGO_HOME` in the sandbox env
4. Also strip `~/.cache`, `~/Library/Caches`, `~/go`, `~/.rustup`, `~/.cargo`, `~/.nvm`, `~/.bun/bin` from `DefaultProfile` (binary resolution now handled by shebang detection + `resolveInnerBinaryPath`)

## Pros

- **Security**: No cross-app cache leakage. Only omac's own subtree is exposed, not other apps' cached data (OAuth tokens, API responses, etc.)
- **Usability**: Tools work out of the box — no config needed. Caches persist across sessions for fast rebuilds.
- **Isolation**: Per-workdir, so different projects don't cross-contaminate.
- **Explicit**: Only 4 well-known tool caches are redirected. Other tools keep their default paths — users add specific cache dirs to their profile when needed.

## Cons

- **Cold start**: First session for a workdir has empty caches (no reuse from system caches). Subsequent sessions are warm.
- **Incomplete coverage**: Only Go, npm, pip, cargo are redirected. Tools without env var overrides (e.g. `gh`, `docker`, custom tools) need manual profile entries.
- **Disk usage**: Per-workdir cache dirs accumulate. No automatic cleanup (could add GC later).
- **Poisoning**: A malicious tool could plant poisoned cache files. Blast radius is limited to omac sessions for that workdir — next session reads its own poisoned cache, not the system's. Much better than poisoning `~/.cache/gh/`.

## Alternatives considered

1. **TMPDIR redirect only**: Redirect all caches to ephemeral `$TMPDIR`. More secure (ephemeral) but cold rebuild every session. Bad usability.
2. **XDG_CACHE_HOME redirect**: Set `XDG_CACHE_HOME` to the per-workdir dir. Catches all XDG-respecting tools automatically, but breaks tools that hardcode `~/.cache/<tool>/` and need existing data. Too broad.
3. **Status quo**: Keep `~/.cache` rw in default profile. Simple but insecure (cross-app leakage).

## Context

Explored during PR #24 (codex+copilot harness support). Reverted to keep that PR focused on harness support. The `SandboxDirs` field (per-harness config/state dirs) stays — it is harness-related. This issue covers the non-harness cache isolation.

NoRiceToday:
## Motivation

Remove default sandbox access to shared host cache roots, preventing cross-application cache leakage and limiting cache poisoning to a selected OMAC trust domain.

Closes #26.

## Implementation Decisions

- Redirect XDG-aware and supported tool caches, including `XDG_CACHE_HOME` and `GOMODCACHE`; unsupported hardcoded paths require explicit profile configuration.
- Use full SHA-256 identities over canonical paths, rather than a short digest.
- Retain read-only Rustup/runtime access while making only the selected cache leaf writable.
- Add explicit cache cleanup and preserve existing profiles; `omac doctor` advises rather than rewriting them.

## Trust And Impact

- Persistent mode isolates host applications and independent OMAC trust domains, but not poisoning between sessions in one domain.
- `--ephemeral-cache` is the stronger opt-in boundary, at cold-start cost.
- Main and linked worktrees receive distinct path-based scopes; multi-directory `serve` intentionally shares one serve scope.
- Host Cargo credentials are not copied; private registries require project-local configuration and an explicit profile-allowed token.
- `--no-sandbox` provides no cache-isolation guarantee.

## Changes Overview

- Add persistent and ephemeral cache scopes with validated cleanup.
- Inject the cache contract through sandbox re-exec and CLI launch paths.
- Add diagnostics, provenance, deterministic runtime/cache E2E tests, and contract documentation.

## Diagram

```mermaid
flowchart LR
  Launch[OMAC launch] --> Scope[Select cache scope]
  Scope --> Env[Inject cache environment]
  Env --> Sandbox[Sandbox re-exec]
  Sandbox --> Tools[Tool runtime]
  Tools --> Cache[Scoped writable cache leaf]
  Cache --> Cleanup[Validated cache cleanup]
```

## Testing

- `go test -race -count=1 ./internal/toolcache`
- `go test -tags=e2e -count=1 -timeout=15m -run ^TestE2ECache ./internal/e2e/`
- `HOME=/tmp go test -count=1 ./internal/cli`

Full host sandbox and Docker verification remain environment-dependent.


nhuelstng (commented):
**What:** Reviewed for correctness and the security boundary of the destructive `cache clear` path. No blocking issues — LGTM pending two optional items inline.

**Verified:**
- `clear`/`clearAll` cannot escape `~/.cache/omac`: symlinked root refused, all mutation via `os.Root` with inode-identity reverification, only `^[0-9a-f]{64}$` entries touched, inner symlinks unlinked in place. Backed by a strong regression suite.
- Cache isolation's real boundary is the filesystem grant (no `~/.cache` in the profile), independently of the env allowlist — proven by `TestE2ECacheEnvironmentOverridesAllowlist` with a restrictive `allow_vars`.
- Build/vet clean; `toolcache` green under `-race`; `cli` tests pass. (The `sandboxrun` escape integration test fails pre-existing on `main` — environment-dependent, unrelated.)

internal/cli/start.go
```diff
@@ -827,6 +785,11 @@ func runLaunch(env *Env, opts launchOpts) int {
 		extra[sandbox.OmacEnvName(m)] = sandbox.OmacTCPEnvValue(m, tcpPort)
 		extra[sandbox.OmacSocketEnvName(m)] = sandbox.OmacEnvValue(m, socketPath)
 	}
+	if cacheScope != nil {
+		for k, v := range toolcache.Environment(cacheScope.Dir, cacheScope.Mode) {
```
nhuelstng: **Optional cleanup — dead env vars.**

**What:** Only `OMAC_CACHE_DIR` / `OMAC_CACHE_MODE` need to be in `extra`; the six tool vars (`XDG_CACHE_HOME`, `GOCACHE`, …) are dead weight here.

**Why:** `sandboxrun.injectedToolCacheEnv` re-derives them and `FilterEnv` overlays `injected` over the inherited value (`env.go:117`), so these are always overridden by identical values. Trimming keeps `run.go` visibly the single authority for the redirect.

**How:** Set only the two `OMAC_CACHE_*` selectors here; let `run.go` produce the tool vars. Same applies to `serve.go:211`. (Non-blocking.)

internal/toolcache/cache.go
```diff
@@ -0,0 +1,135 @@
...
+			return err
+		}
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return fmt.Errorf("cache directory %q is a symlink", path)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("cache path %q is not a directory", path)
+	}
+	return os.Chmod(path, 0o700)
```
nhuelstng: **Optional hardening — path-based chmod follows symlinks.**

**What:** `ensurePrivateDir` ends with `os.Chmod(path, 0o700)`, which follows symlinks; the rest of the package uses fd + inode-identity discipline.

**Why:** If `~/.cache/omac` (or `.locks`) is swapped for a symlink in the Lstat→Chmod window, this chmods the *target* to `0700`. Low severity — the precondition is an attacker who can already write to `~/.cache` — but it's the one spot breaking the `os.Root`/`fchmod` pattern the destructive path uses everywhere else.

**How:** Open the dir and `fchmod` the descriptor (mirror `ensureRootPrivateDir` in `lock.go`), or note the accepted risk. (Non-blocking.)

nhuelstng:
this is fine for me once the tests are green :)

internal/cli/start.go
NoRiceToday: Fixed. start.go and serve.go now inject only OMAC_CACHE_DIR / OMAC_CACHE_MODE; the six tool redirects are left to sandboxrun.run.go (injectedToolCacheEnv), which is the single authority.

internal/toolcache/cache.go
NoRiceToday: Fixed. ensurePrivateDir now opens the directory descriptor and fchmods it, mirroring the os.Root pattern in ensureRootPrivateDir — no path-based chmod that could follow a swapped symlink.

NoRiceToday:
FYI: to speed up the testing, I build https://github.com/TNG/oh-my-agentic-coder/pull/118 and rebased this one on top of that one. So we should first merge the testing pr before we merge this one

NoRiceToday:
Ready for review @nhuelstng

nhuelstng (approved):

PR #111, opened 2026-07-16
fix(sandbox): ship default env allowlist; strip ambient host env (#102)

### LINKED ISSUE #102: Sandbox profiles pass ambient environment variables by default
## Observed behavior

The built-in sandbox default profile ships with an empty `environment.allow_vars`. `FilterEnv` treats an empty allowlist as "allow every inherited variable except the small dangerous-variable blocklist" (`internal/sandboxprofile/env.go:96`, `internal/sandboxprofile/profile.go:208`).

As a result the sandboxed agent inherits arbitrary ambient parent-process environment variables: credentials (`AWS_SECRET_ACCESS_KEY`-style secrets), capability pointers (`DOCKER_HOST`, `SSH_AUTH_SOCK`), proxy settings, and harness metadata. The blocklist only drops a fixed set of dangerous interpreter/loader variables (`BASH_ENV`, `LD_*`, `DYLD_*`, `NODE_OPTIONS`, `OP_SESSION_*`, etc.); everything else passes through.

The compiled-in `DefaultProfile()` (`internal/sandboxprofile/resolve.go:23`) omits an allowlist, and the scaffolded `default.json` writes `"environment": {}`. The e2e allowance spec already uses an explicit allowlist (`internal/e2e/allowance.go:75`) — the default profile does not.

This contradicts the documented trust model: "the sandbox never sees any secret, any host env var that isn't explicitly forwarded" (`oh-my-agentic-coder.md` §16.9).

## Risk

- Ambient secrets in the launching shell (API tokens, cloud credentials) leak into the sandbox unfiltered.
- Capability pointers (`DOCKER_HOST`, `SSH_AUTH_SOCK`) become visible; when the referenced socket/file is also granted by the profile, the agent gains host-level access, not just knowledge of the path.
- Harness metadata and proxy configuration are exposed without explicit opt-in.
- The product ships insecure-by-default; users must know to add an allowlist.

## Proposed fix

Ship a default `environment.allow_vars` allowlist with the minimal set the runtime needs, e.g.:

```json
"environment": {
  "allow_vars": [
    "OMAC_*",
    "HOME",
    "PATH",
    "PWD",
    "TMPDIR",
    "LANG",
    "LC_*",
    "TERM",
    "SHELL",
    "XDG_CONFIG_HOME",
    "XDG_DATA_HOME",
    "XDG_STATE_HOME",
    "NPM_CONFIG_PREFIX"
  ]
}
```

Harness-specific required variables (e.g. `SKAINET_TOKEN` for opencode/codex/copilot, `ANTHROPIC_AUTH_TOKEN` for claude-code) should be merged in at launch from `Harness` metadata, mirroring the e2e allowance spec.

Injected variables (`HTTP_PROXY`/`HTTPS_PROXY`, `OMAC_*`) already bypass the allowlist via the `injected` overlay and remain unaffected.

## Acceptance

- Default profile (compiled-in and scaffolded) contains an explicit `allow_vars` list.
- Unlisted inherited variables are absent from the child env.
- Regression test asserts an ambient secret not in the allowlist is stripped (the existing `TestFilterEnv` no-allowlist case documents the current leak; add the inverse).
- Migration note for users with existing scaffolded profiles.

nhuelstng:
## What

Harden the sandbox environment so it forwards only what is explicitly needed, and make a misconfigured (empty) allowlist fail **safe** rather than leak or break.

- **Hardened default profile.** `DefaultProfile()` (compiled-in → scaffolded `default.json`) ships an explicit `environment.allow_vars` — the operational minimum: `OMAC_*`, `HOME`, `PATH`, `PWD`, `TMPDIR`, `LANG`, `LC_*`, `TERM`, `SHELL`, `USER`, `LOGNAME`, `TZ`, `EDITOR`, `VISUAL`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `XDG_STATE_HOME`, `NPM_CONFIG_PREFIX`.
- **Targeted per-harness auth forwarding.** Each *single-provider* `Harness` declares the auth vars it unambiguously needs (claude-code → `ANTHROPIC_*`, codex → `OPENAI_*`, copilot → `GITHUB_TOKEN`/`GH_TOKEN` + `COPILOT_CUSTOM_INSTRUCTIONS_DIRS`); omac injects them for the *selected* harness at launch via a new additive `--allow-env` flag on `omac sandbox run`. **Multi-provider harnesses (opencode, pi) forward nothing** — auto-pushing a grab-bag of unrelated third-party keys overshoots the trust model; those setups declare the specific key they use in `allow_vars`.
- **Empty allowlist fails closed, with a heads-up.** On `omac start`/`serve`, a profile with an empty `allow_vars` no longer inherits every ambient var. omac warns, pauses briefly (~2s), and forwards **only** the operational minimum (no auto-auth). The harness starts but authenticates only once the profile declares what it needs.
- **Two diagnostics.** `omac provenance --check` flags an empty `allow_vars` (MEDIUM); `omac doctor` warns for every native profile with an empty `allow_vars`, with impact + remediation.

## Why

Closes #102. An empty `allow_vars` made `FilterEnv` inherit **every** ambient parent-process variable except the small danger blocklist: cloud/CI secrets, capability pointers (`DOCKER_HOST`, `SSH_AUTH_SOCK`), proxy config, harness metadata. This contradicted the documented trust model — *"the sandbox never sees any host env var that isn't explicitly forwarded"* (`oh-my-agentic-coder.md` §16, README §What the sandbox can see).

Two follow-on risks surfaced during review and had to be handled for the model to stay coherent:
- The launch-time `--allow-env` injection, appended to an **empty** profile, would flip it from inherit-all into a restrictive list of just those names — silently stripping `HOME`/`PATH`/tokens and breaking the harness. Hence fail-closed-to-operational-minimum **plus a visible warning**, rather than a silent break.
- Blindly forwarding every provider key a multi-provider harness *could* read overshoots "forward only what's needed"; forwarding is now targeted, and custom/unknown provider tokens are declared per-profile.

## How

- `sandboxprofile.DefaultAllowVars()` defines the operational minimum; `DefaultProfile()` sets it, so both the compiled-in default and the scaffolded `default.json` carry an explicit allowlist.
- `--allow-env` is additive into `Environment.AllowVars` (`Merge`), injected per selected harness at launch (mirroring how `SandboxDirs` become `--allow` flags). Injection is gated on a **template-anchored** check (`profileRunsNativeSandbox`: the profile's command template begins with `{{self}} sandbox run`) rather than an unanchored token scan — so a `nono` profile whose inner command merely contains `sandbox`/`run` tokens is never misclassified, and `{{self}}` resolving to an absolute `os.Executable()` path is irrelevant.
- `forwardHarnessEnv` resolves the referenced profile read-only; on an empty `allow_vars` it warns, sleeps (`emptyAllowVarsWarnDelay`, 2s), and seeds **only** `DefaultAllowVars()` — deliberately *not* the harness auth vars (omac does not silently push provider credentials into a misconfigured sandbox). Non-empty profiles inject the harness's auth vars as before.
- Injected vars (`HTTP(S)_PROXY`, `OMAC_*`) still bypass the allowlist via the injected overlay — unchanged.
- Users on other/custom providers (e.g. a TNG-style `SKAINET_*`) add the var to the profile's `allow_vars`; documented in the README migration note, which also covers the `allow_vars: ["*"]` inherit-everything escape hatch.

## Verification

- `go test ./...` green; `go build -tags e2e ./...`, `gofmt`, `go vet` clean. Full CI green including the e2e **Cache integration** and **model-free** jobs.
- Tests:
  - `FilterEnv` strips an ambient secret under the default allowlist; default profile excludes `AWS_SECRET_ACCESS_KEY`; golden-list pins the exact `DefaultAllowVars()` membership.
  - `--allow-env` parse + additive merge (no input mutation); merged launch-env asserted end-to-end.
  - Template-anchored native detection exercised against real `Expand` output; a `nono` profile with inner `sandbox`/`run` tokens is left untouched.
  - Empty profile at launch forwards the operational minimum, does **not** auto-forward auth vars, and warns; a non-empty profile still injects.
  - `omac doctor` warns on empty `allow_vars` (impact/remediation) and is silent on an explicit list.
  - Empty `allow_vars` produces a MEDIUM audit finding; `allow_vars: ["*"]` suppresses it.
  - Auth-forward policy: opencode/pi forward nothing; claude-code/codex/copilot keep their targeted keys.
  - e2e cache-isolation fixture declares `allow_vars: ["*"]` (it exercises cache-dir isolation, not env allowlisting).

## Acceptance criteria

- [x] Default profile (compiled-in and scaffolded) contains an explicit `allow_vars` list.
- [x] Unlisted inherited variables are absent from the child env.
- [x] Regression test asserts an ambient secret not in the allowlist is stripped.
- [x] Migration note for users with existing scaffolded profiles.
- [x] An empty/misconfigured profile fails **closed** (operational minimum) with a launch warning — not a silent leak or a broken launch.
- [x] Auth forwarding is targeted per harness; multi-provider harnesses declare their key explicitly.

### DIFF
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -398,6 +398,10 @@ func runServe(args []string, env *Env) int {
 		if cacheScope != nil {
 			argv = injectSandboxFlag(argv, "--allow", cacheScope.Dir)
 		}
+		// Forward the selected harness's auth env vars through the
+		// default profile's restrictive allow_vars filter. (Control-plane
+		// port and harness runtime dirs are granted inside sandboxServeArgv.)
+		argv = injectSandboxEnvAllow(argv, harness.SandboxEnvAllow)
 		// Pass the resolved audit path to `omac sandbox run` so its
 		// network-filter subprocess appends net.decision events to the
 		// same persistent log. Inherit the parent's run_id + mode so the
@@ -567,6 +571,41 @@ func injectSandboxDirs(argv []string, dirs []string) []string {
 		argv = injectSandboxFlag(argv, "--allow", d)
 	}
 	return argv
+}
+
+// injectSandboxEnvAllow splices --allow-env flags for each harness-declared
+// auth env var into the sandbox argv, so they survive the default profile's
+// restrictive allow_vars filter. Only the selected harness's vars are added.
+//
+// --allow-env is understood only by omac's native `sandbox run` backend
+// (where FilterEnv applies the allowlist); other backends (nono) do their
+// own env filtering via their own profile, so injecting the flag there
+// would be meaningless and could fail on an unknown flag. Guarded by
+// argvRunsNativeSandbox. Empty/nil is a no-op.
+func injectSandboxEnvAllow(argv []string, names []string) []string {
+	if !argvRunsNativeSandbox(argv) {
+		return argv
+	}
+	for _, n := range names {
+		if n == "" {
+			continue
+		}
+		argv = injectSandboxFlag(argv, "--allow-env", n)
+	}
+	return argv
+}
+
+// argvRunsNativeSandbox reports whether argv invokes omac's native
+// sandbox supervisor (`omac sandbox run …`), identified by consecutive
+// "sandbox" "run" tokens. This is the only backend that parses the
+// launch-injected sandbox flags omac itself defines (e.g. --allow-env).
+func argvRunsNativeSandbox(argv []string) bool {
+	for i := 0; i+1 < len(argv); i++ {
+		if argv[i] == "sandbox" && argv[i+1] == "run" {
+			return true
+		}
+	}
+	return false
 }
 
 // injectSandboxFlag splices a sandbox flag (with optional value; pass
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -27,6 +27,7 @@ import (
 	"github.com/tngtech/oh-my-agentic-coder/internal/opencodestate"
 	"github.com/tngtech/oh-my-agentic-coder/internal/registry"
 	"github.com/tngtech/oh-my-agentic-coder/internal/sandbox"
+	"github.com/tngtech/oh-my-agentic-coder/internal/sandboxprofile"
 	"github.com/tngtech/oh-my-agentic-coder/internal/secrets"
 	"github.com/tngtech/oh-my-agentic-coder/internal/skillconfig"
 	"github.com/tngtech/oh-my-agentic-coder/internal/skillsource"
@@ -398,10 +399,10 @@ func runServe(args []string, env *Env) int {
 		if cacheScope != nil {
 			argv = injectSandboxFlag(argv, "--allow", cacheScope.Dir)
 		}
-		// Forward the selected harness's auth env vars through the
-		// default profile's restrictive allow_vars filter. (Control-plane
-		// port and harness runtime dirs are granted inside sandboxServeArgv.)
-		argv = injectSandboxEnvAllow(argv, harness.SandboxEnvAllow, prof)
+		// Forward the selected harness's auth env vars through the default
+		// profile's restrictive allow_vars filter. (Control-plane port and
+		// harness runtime dirs are granted inside sandboxServeArgv.)
+		argv = forwardHarnessEnv(env, argv, harness, prof, profName)
 		// Pass the resolved audit path to `omac sandbox run` so its
 		// network-filter subprocess appends net.decision events to the
 		// same persistent log. Inherit the parent's run_id + mode so the
@@ -571,6 +572,53 @@ func injectSandboxDirs(argv []string, dirs []string) []string {
 		argv = injectSandboxFlag(argv, "--allow", d)
 	}
 	return argv
+}
+
+// emptyAllowVarsWarnDelay is how long omac pauses after warning about an
+// empty-allow_vars profile at launch, so the message is seen before the
+// harness UI takes over. Overridable in tests.
+var emptyAllowVarsWarnDelay = 2 * time.Second
+
+// forwardHarnessEnv forwards the selected harness's auth env vars into the
+// sandbox via --allow-env. If the resolved sandbox profile has an EMPTY
+// environment.allow_vars, leaving it empty would inherit every ambient host
+// var (the pre-#102 leak). Rather than either leak or break, omac fails
+// closed: it seeds ONLY the operational minimum (sandboxprofile.DefaultAllowVars)
+// so the empty inherit-all list becomes a restrictive allowlist — ambient
+// secrets are stripped while HOME/PATH/locale keep the harness runnable.
+// Provider-auth vars are deliberately NOT auto-forwarded here: an empty
+// profile is a misconfiguration, and omac does not silently push provider
+// credentials into the sandbox to paper over it. It warns that this differs
+// from the old inherit-everything behavior and pauses so the message is seen.
+func forwardHarnessEnv(env *Env, argv []string, harness config.Harness, prof config.SandboxProfile, profName string) []string {
+	if empty, ok := sandboxProfileAllowVarsEmpty(prof); ok && empty {
+		fmt.Fprintf(env.Stderr, "omac: sandbox profile %q has an empty environment.allow_vars.\n", profName)
+		fmt.Fprintln(env.Stderr, "      Forwarding only the operational minimum (HOME, PATH, TERM, locale, …).")
+		fmt.Fprintln(env.Stderr, "      No provider-auth or other ambient env vars are passed through (previously every")
+		fmt.Fprintln(env.Stderr, "      var was inherited). The harness starts but will not authenticate until you add")
+		fmt.Fprintln(env.Stderr, `      the vars it needs to allow_vars (see "omac doctor"), or set allow_vars: ["*"].`)
+		fmt.Fprintln(env.Stderr, "      Continuing shortly…")
+		time.Sleep(emptyAllowVarsWarnDelay)
+		// Seed only the operational minimum; do NOT auto-forward auth vars.
+		return injectSandboxEnvAllow(argv, sandboxprofile.DefaultAllowVars(), prof)
+	}
+	return injectSandboxEnvAllow(argv, harness.SandboxEnvAllow, prof)
+}
+
+// sandboxProfileAllowVarsEmpty reports whether the native sandbox profile
+// referenced by prof resolves to an empty environment.allow_vars. ok is false
+// when prof is not an inspectable native ({{self}} sandbox run) profile or the
+// referenced profile cannot be resolved read-only.
+func sandboxProfileAllowVarsEmpty(prof config.SandboxProfile) (empty bool, ok bool) {
+	ref, isNative := inspectBuiltinProfileRef(prof.Command)
+	if !isNative {
+		return false, false
+	}
+	sp, _, err := sandboxprofile.ResolveReadOnly(ref)
+	if err != nil {
+		return false, false
+	}
+	return len(sp.Environment.AllowVars) == 0, true
 }
 
 // injectSandboxEnvAllow splices --allow-env flags for each harness-declared
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -401,7 +401,7 @@ func runServe(args []string, env *Env) int {
 		// Forward the selected harness's auth env vars through the
 		// default profile's restrictive allow_vars filter. (Control-plane
 		// port and harness runtime dirs are granted inside sandboxServeArgv.)
-		argv = injectSandboxEnvAllow(argv, harness.SandboxEnvAllow)
+		argv = injectSandboxEnvAllow(argv, harness.SandboxEnvAllow, prof)
 		// Pass the resolved audit path to `omac sandbox run` so its
 		// network-filter subprocess appends net.decision events to the
 		// same persistent log. Inherit the parent's run_id + mode so the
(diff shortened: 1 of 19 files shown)


internal/cli/serve.go
```diff
@@ -490,6 +493,41 @@ func injectSandboxDirs(argv []string, dirs []string) []string {
...
+// sandbox supervisor (`omac sandbox run …`), identified by consecutive
+// "sandbox" "run" tokens. This is the only backend that parses the
+// launch-injected sandbox flags omac itself defines (e.g. --allow-env).
+func argvRunsNativeSandbox(argv []string) bool {
+	for i := 0; i+1 < len(argv); i++ {
+		if argv[i] == "sandbox" && argv[i+1] == "run" {
+			return true
+		}
+	}
+	return false
```
NoRiceToday: The `argv[i] == "sandbox" && argv[i+1] == "run"` scan is unanchored on `argv[0]`. The `builtin` profile template (`internal/config/launcher.go:117`) emits `["{{self}}", "sandbox", "run", …]` where `{{self}}` resolves to `os.Executable()` — not the literal `"omac"` the test uses (`serve_test.go:223`). The detector passes today only because it ignores `argv[0]`, so the fragility is invisible to the test suite.

Two concrete failure modes: (a) a future profile that reorders subcommand tokens (e.g. `omac sandbox --profile x run`) silently drops every harness auth var; (b) a user-authored profile whose inner_cmd contains `["sandbox","run",...]` would match and get `--allow-env` flags spliced into the wrong position.

Suggestion: pass the selected `SandboxProfile`/`profName` into `injectSandboxEnvAllow` and branch on profile name (`builtin` → inject; `nono*` → skip), or anchor on `argv[0] == os.Executable()`. Add a test that uses the real `Expand` output, not a hand-rolled `"omac"` argv.

[maintainability]

internal/sandboxprofile/env.go
```diff
@@ -41,6 +41,38 @@ var dangerousEnvPrefixes = []string{
...
+		"LANG",
+		"LC_*",
+		"TERM",
+		"SHELL",
+		"XDG_CONFIG_HOME",
+		"XDG_DATA_HOME",
+		"XDG_STATE_HOME",
+		"NPM_CONFIG_PREFIX",
+	}
+}
```
NoRiceToday, line 84: `DefaultAllowVars` is missing vars the operational minimum usually needs:

- `USER` / `LOGNAME` — `FacadeConfig.BaseEnvPassthrough` (`internal/config/launcher.go:248`) explicitly forwards `USER` to the facade; the sandbox allowlist doesn't, so the inner process loses it. git/npm/ssh default-identity lookups break.
- `TZ` — date formatting in tool output, log lines, test snapshots silently fall back to UTC.
- `EDITOR` / `VISUAL` — harnesses that shell out to an editor (git commit message) break.

Also: `TestDefaultProfileEnvAllowlist` (`profile_test.go:559`) only checks `OMAC_*`, `HOME`, `PATH` are present and `AWS_SECRET_ACCESS_KEY` is absent — it does NOT assert the full membership, so future drift (e.g. someone removing `SHELL`) isn't caught. Add a golden-list test asserting the exact `DefaultAllowVars()` slice.

[bug] [missing-test]

internal/sandboxprofile/env.go
```diff
@@ -41,6 +41,38 @@ var dangerousEnvPrefixes = []string{
...
+//
+// Entries are exact names or trailing-* prefixes (see envVarAllowed).
+// Injected variables (HTTP(S)_PROXY, OMAC_* skill bases) bypass the
+// allowlist entirely via the injected overlay in FilterEnv, but OMAC_*
+// is listed too because omac sets several OMAC_* vars in its own process
+// environment before exec (see internal/cli/start.go), and those reach
+// the child through inheritance rather than the injected overlay.
+func DefaultAllowVars() []string {
+	return []string{
+		"OMAC_*",
```
NoRiceToday, line 65: Pattern-matching note (for the whole `DefaultAllowVars` list, anchored here on `OMAC_*`):

`envVarAllowed` works correctly for `OMAC_*` / `LC_*` (trailing-`*` prefix, anchored at key start — `FOO_OMAC_*` does not match, confirmed). One gap: the matcher accepts any pattern shape, so a typo like `OMAC` (no `*`) silently matches only the literal `OMAC`, not `OMAC_BASE`. Not a bug, but worth either validating the pattern shape in `Validate()` (`profile.go:259`) or documenting the trailing-`*`-only contract next to the `Environment.AllowVars` field so users authoring `default.json` know what is accepted.

Also note: `OMAC_*` is a wide gate. Any skill sidecar that sets `OMAC_EVIL` in the host env would pass. The trust model accepts this, but it is worth a one-line note in the README that `OMAC_*` is a broad prefix and sidecars must not set arbitrary `OMAC_*` vars in the host env.

[maintainability] [docs-drift]

internal/profileaudit/check.go
```diff
@@ -21,10 +21,30 @@ func Check(profile *sandboxprofile.Profile) []Finding {
...
+func checkEnvironment(profile *sandboxprofile.Profile) []Finding {
+	if len(profile.Environment.AllowVars) > 0 {
+		return nil
+	}
+	return []Finding{{
+		Severity: SeverityMedium,
+		Category: CatEnvironment,
+		Field:    "environment.allow_vars",
+		Value:    "(empty)",
+		Message:  "empty allowlist inherits all ambient env vars except the danger blocklist; add an explicit allow_vars list",
```
NoRiceToday, line 44: `checkEnvironment` fires MEDIUM on any empty `allow_vars`, including intentionally-empty user profiles (debug, `no-sandbox-debug`). There is no documented escape hatch, though `allow_vars: ["*"]` already suppresses the finding (length > 0) AND is safe because the blocklist still wins at `env.go:143-145`.

Two asks:

1. Document `allow_vars: ["*"]` as the explicit "inherit everything" escape hatch in the README migration note (`README.md:556-560`).
2. The README migration note says existing empty `default.json` files should "delete the file to let omac re-scaffold it" — but does not mention that hand-edited profiles would be lost, nor that `Resolve` only scaffolds on first-miss (`resolve.go:113-126`), so existing empty files are NOT auto-migrated. Suggest: "Existing empty-environment `default.json` files are not auto-migrated. Run `omac provenance --check` (now MEDIUM). Either back up + delete to re-scaffold, add an `allow_vars` list (see `DefaultAllowVars()` in `internal/sandboxprofile/env.go`), or set `allow_vars: ["*"]` to restore the pre-#102 behavior."

[docs-drift] [decision]

internal/sandboxprofile/profile_test.go
```diff
@@ -517,6 +540,45 @@ func TestFilterEnv(t *testing.T) {
...
+	for k, seen := range want {
+		if !seen {
+			t.Errorf("default allowlist missing %q", k)
+		}
+	}
+	// A known ambient secret must NOT be allowed by the default list.
+	if envVarAllowed("AWS_SECRET_ACCESS_KEY", p.Environment.AllowVars) {
+		t.Error("default allowlist must not permit AWS_SECRET_ACCESS_KEY")
+	}
 }
```
NoRiceToday, line 706: Coverage gaps in the new tests:

- No golden-list test for `DefaultAllowVars()` membership — `TestDefaultProfileEnvAllowlist` (line 562) only checks `OMAC_*`, `HOME`, `PATH` are present and `AWS_SECRET_ACCESS_KEY` is absent, so future drift (e.g. someone removing `SHELL` or `USER`) is not caught. See comment on `env.go:58`.
- `TestInjectSandboxEnvAllow` (`serve_test.go:223`) uses a hand-rolled `["omac","sandbox","run",...]` argv instead of the real `Expand` output — masks the fragility in `argvRunsNativeSandbox` (see comment on `serve.go:522`).
- No test asserting `allow_vars: ["*"]` suppresses the audit finding (see comment on `check.go:35`).
- No end-to-end test asserting the merged env (`DefaultAllowVars() + harness.SandboxEnvAllow` + injected `HTTP(S)_PROXY`/`OMAC_*` overlay) produces the exact expected child env. The closest is the `e2e` build tag, not run in default CI.

[missing-test]

NoRiceToday:
Review complete — requesting changes. Five inline comments posted (findings #2–#5 + a consolidated test-coverage note). Summary of asks, ordered by priority:

**Should fix now:**

1. **`argvRunsNativeSandbox` fragility** (`internal/cli/serve.go:522`) — the `"sandbox" "run"` token scan is unanchored on `argv[0]`; the `builtin` template emits `["{{self}}", "sandbox", "run", …]` where `{{self}}` is `os.Executable()`, not the literal `"omac"` the test uses. Pass the profile name into `injectSandboxEnvAllow` instead of string-sniffing the argv, and add a test using the real `Expand` output.

2. **`DefaultAllowVars` membership drift** (`internal/sandboxprofile/env.go:58`) — missing `USER`/`LOGNAME` (which `FacadeConfig.BaseEnvPassthrough` *does* forward to the facade, so the inner process loses it), `TZ`, and arguably `EDITOR`/`VISUAL`. Add a golden-list test asserting the exact slice so future drift is caught.

3. **Docs — trust model + migration note** (`internal/profileaudit/check.go:35`, `README.md:556-560`) — document `allow_vars: ["*"]` as the explicit "inherit everything" escape hatch (audit-safe because the blocklist still wins at `env.go:143-145`). Rewrite the migration note to state that existing empty-`environment` `default.json` files are NOT auto-migrated (`Resolve` only scaffolds on first-miss) and that hand-edited profiles should be backed up before deletion.

**Nice to have:**

4. **Pattern-shape validation** (`internal/sandboxprofile/env.go:107`) — `envVarAllowed` accepts any pattern, so a typo like `OMAC` (no `*`) silently matches only the literal `OMAC`. Either validate the pattern shape in `Validate()` or document the trailing-`*`-only contract next to `Environment.AllowVars`. Also worth a README note that `OMAC_*` is a broad prefix and sidecars must not set arbitrary `OMAC_*` vars in the host env.

5. **Test coverage** (`internal/sandboxprofile/profile_test.go`) — add the golden-list test, the `["*"]`-suppresses-audit test, and an end-to-end merged-env test (the `e2e` build tag is not run in default CI).

Native `builtin`/`sandbox run` path is solid — scaffold, additive `--allow-env` merge, no input mutation, blocklist-beats-allowlist, and the MEDIUM audit finding are all genuinely tested. No blockers; the above are hardening + docs + test gaps.

NoRiceToday (commented):
Sieht soweit gut aus, habe paar Verbesserungsvorschläge hinterlassen. Rein funktional, hab ich es bereits selbst getestet 👍

PR #106, opened 2026-07-16
docs: add collaboration guide and issue/PR spec

NoRiceToday:
## What
- Adds `COLLABORATION.md` at repo root: the repo's collaboration workflow (issue-tracked work, self-assignment, small feat/fix branches, explicit Approve/Request-changes reviews, re-request review after feedback, no merge without approval, resolve-comments-as-you-go rule).
- Adds a concise agent-first **spec for issues and PRs** (What/Why/How/Verification, Non-goals, conventional-commit PR titles with scope, `Closes #NN` linking).
- Adds `.github/ISSUE_TEMPLATE/issue.yml` and `.github/PULL_REQUEST_TEMPLATE.md` enforcing the structure at creation time.
- Resolves the **specs** open question: keep specs inline in the issue body; persist architecture/pattern decisions via `docs/`/`AGENTS.md`/ADR, not a `specs/` tree.
- Adds an `AGENTS.md` pointer (with `CLAUDE.md` symlink) directing agents to read and adhere to `COLLABORATION.md` before starting any work.

## Why
Humans and agents contribute together; we need one short reference so everyone follows the same workflow and so new issues/PRs follow a consistent, agent-friendly structure. Tracked in #105.

## How
- Researched ~12 existing issues and ~10 PRs to extract the de-facto conventions (PR titles use Conventional Commits with scope; PR bodies follow What/Why/How/Verification; substantive issues carry Non-goals; linking keywords mixed so we pick `Closes`/`Refs`).
- Kept the doc concise (~125 lines), mirroring existing conventions rather than inventing new ones.
- Turned the structure into templates so the convention is enforced at creation time, not just documented.

## Verification
- `git log --oneline -1` -> `docs: add templates, resolve specs question, review-comment rules`
- `git diff --stat main..HEAD` -> `COLLABORATION.md`, `AGENTS.md`, `CLAUDE.md`, `.github/ISSUE_TEMPLATE/issue.yml`, `.github/PULL_REQUEST_TEMPLATE.md`
- Issue created and assigned: #105
- Specs question resolved per discussion in #105.

Closes #105

🤖 Generated with opencode

PR #112, opened 2026-07-16
feat(sandbox): upstream proxy chaining for corporate environments

GevatterGaul:
## Problem

In corporate setups, the LLM API is only reachable through a corporate proxy
(`HTTPS_PROXY` set in the host environment). omac's built-in sandbox runs its
own filtering HTTP CONNECT proxy and unconditionally overwrites the child's
`HTTP_PROXY`/`HTTPS_PROXY` env vars to point at the omac proxy. The omac proxy
then dialed upstream targets directly — with no support for chaining through a
corporate proxy. This made omac unusable behind corporate firewalls.

"Enterprise upstream-proxy chaining" was explicitly listed as a non-goal in the
original native-sandbox design. This PR implements it.

## Solution

Introduces a `Dialer` abstraction in `internal/netproxy` that replaces the
hardcoded `dialPinned` call path. Two implementations:

- **`directDialer`** — the existing behavior: resolve DNS locally, pin IPs
  (anti-DNS-rebinding), dial directly.
- **`upstreamProxyDialer`** — dial the upstream corporate proxy, send
  `CONNECT host:port`, parse the `200` response, and splice the raw tunnel.
  Passes the **hostname** (not pre-resolved IPs) so the corporate proxy can do
  its own DNS (internal hostnames, PAC routing).

The omac filter (allow/deny domains, metadata hard-deny, interactive prompts)
**always runs first** — the upstream proxy is purely a transport underneath.
`NO_PROXY` only selects between chained vs direct transport; it never bypasses
the security filter.

### Config

- **Auto-detection (zero-config):** `HTTPS_PROXY` / `HTTP_PROXY` / `NO_PROXY`
  are read from the host environment. Corporate users with proxy env vars
  already set get chaining automatically.
- **Profile override:** `network.upstream_proxy` and `network.no_proxy` in the
  sandbox profile JSON override the env vars.
- **Auth:** Basic auth from the proxy URL
  (`http://user:pass@proxy:8080`). NTLM/Kerberos not supported directly —
  documented workaround: run `cntlm` or `px` as a local bridge.
- **Errors:** `502 Bad Gateway` with `X-Omac-Sandbox: upstream-error` header,
  naming the upstream proxy host (never credentials).

## Files

| File | Change |
|---|---|
| `internal/netproxy/dialer.go` *(new)* | `Dialer` interface, `directDialer`, `upstreamProxyDialer`, `UpstreamError`, `ProxyAuthenticator` interface, `bufferedConn`, `hostMatchesNoProxy` |
| `internal/netproxy/dialer_test.go` *(new)* | 20 tests: CONNECT tunneling, Proxy-Auth emission, UpstreamError on 407/503, NO_PROXY bypass, buffered conn, dial failure, context cancel |
| `internal/netproxy/server.go` | `dialer Dialer` field, `NewServer` accepts `Dialer`, `handleConnect`/`handleForward` use `s.dialer.DialTunnel`, `ProxyAuthenticator` check for absolute-URI + upstream auth, 502 with `X-Omac-Sandbox: upstream-error` |
| `internal/netproxy/server_test.go` | 5 new server-level tests: end-to-end CONNECT through upstream, filter-deny-before-chain, 502 on upstream failure, 502 on dial failure, forward with auth |
| `internal/sandboxprofile/profile.go` | `network.upstream_proxy` + `network.no_proxy` fields + validation |
| `internal/sandboxprofile/profile_test.go` | 4 new test functions (10 sub-cases) covering all validation rules |
| `internal/sandboxrun/run.go` | `resolveUpstreamProxy` — resolves profile→env, constructs the right `Dialer` |
| `README.md` | "Corporate proxy" section under Configuration |
| `openspec/changes/native-sandbox/design.md` | Removed "enterprise upstream-proxy chaining" from non-goals |

## Verification

- `go build ./...` — clean
- `go vet ./internal/netproxy/ ./internal/sandboxprofile/ ./internal/sandboxrun/` — clean
- `go test -race ./internal/netproxy/ -count=1` — **48 tests pass** (25 new + 23 existing)
- `go test ./internal/sandboxprofile/ -count=1` — **29 tests pass** (4 new + 25 existing)

## Test coverage

**Dialer unit tests (20):** direct dialer regression, CONNECT tunneling with
hostname (not IP), Proxy-Authorization present/absent based on URL userinfo,
UpstreamError on 407/503/502-with-body, NO_PROXY bypass (match→direct,
non-match→chained), bufferedConn pipelined bytes, dial failure attribution,
context cancellation, filter-enforcement layering boundary.

**Server-level tests (5):** end-to-end CONNECT through upstream proxy chain
(data flows through), filter denies before upstream dial (0 upstream
connections on deny), 502 with `upstream-error` header on upstream rejection,
502 on upstream dial failure, plain-HTTP forward through authenticated upstream.

## Security considerations

- The omac session token (`Proxy-Authorization: Basic omac:<token>`) is **always
  stripped** before forwarding to the upstream proxy, then replaced with the
  upstream's credentials if configured.
- `UpstreamError` carries only `ProxyHost` (host:port, no userinfo),
  `StatusLine`, and `Err` — never credentials.
- The log output uses `proxyURL.Host` only — no userinfo is ever logged.
- Anti-DNS-rebinding is preserved for the direct path (resolve+pin). For the
  chained path, DNS is delegated to the upstream proxy (unavoidable — the filter
  operates on the hostname the child requested, which is the security boundary).

### DIFF
--- a/internal/sandboxprofile/profile.go
+++ b/internal/sandboxprofile/profile.go
@@ -10,13 +10,15 @@
 //	  "workdir": { "access": "readwrite" },
 //	  "filesystem": { "allow": [...], "read": [...], "write": [...],
 //	                  "deny": [...], "override_deny": [...] },
-//	  "network": { "mode": "filtered", "allow_domain": [...],
+//	 "network": { "mode": "filtered", "allow_domain": [...],
 //	               "deny_domain": [...], "listen_port": [...],
 //	               "allow_tcp_connect": [...], "open_port": [...],
 //	               "network_prompt": { "enabled": true,
 //	                                   "prompt_timeout_secs": 60,
 //	                                   "on_unavailable": "deny" },
-//	               "enforcement": "kernel" },
+//	               "enforcement": "kernel",
+//	               "upstream_proxy": "http://proxy:8080",
+//	               "no_proxy": ["internal.corp"] },
 //	  "environment": { "allow_vars": [...] }
 //	}
 //
@@ -28,6 +30,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"path/filepath"
 	"strings"
 )
@@ -132,6 +135,18 @@ type Network struct {
 	// Enforcement is kernel (default) or env-only (Linux escape hatch
 	// for kernels without Landlock ABI >= 4; advisory only).
 	Enforcement string `json:"enforcement,omitempty"`
+	// UpstreamProxy is an optional http:// or https:// URL pointing to a
+	// proxy that all outbound traffic must be routed through.  When set,
+	// the sandbox must enforce that every outbound connection goes through
+	// this proxy.
+	UpstreamProxy string `json:"upstream_proxy,omitempty"`
+	// NoProxy is a list of hostnames or parent domains that must bypass the
+	// upstream proxy and be dialed directly. Matching is case-insensitive:
+	// an entry matches that exact host or any subdomain of it (e.g.
+	// "example.com" also matches "api.example.com"). CIDR ranges and
+	// leading-dot forms (".example.com") are not supported. Empty entries
+	// are rejected.
+	NoProxy []string `json:"no_proxy,omitempty"`
 }
 
 // NetworkPrompt mirrors nono's network_prompt block.
@@ -255,6 +270,17 @@ func (p *Profile) Validate() error {
 		}
 		if np.PromptTimeoutSecs < 0 {
 			return fmt.Errorf("sandbox profile: network.network_prompt.prompt_timeout_secs must be >= 0")
+		}
+	}
+	if p.Network.UpstreamProxy != "" {
+		u, err := url.Parse(p.Network.UpstreamProxy)
+		if err != nil || u.Scheme != "http" || u.Host == "" {
+			return fmt.Errorf("sandbox profile: invalid network.upstream_proxy %q (want http://host[:port])", p.Network.UpstreamProxy)
+		}
+	}
+	for _, entry := range p.Network.NoProxy {
+		if strings.TrimSpace(entry) == "" {
+			return fmt.Errorf("sandbox profile: network.no_proxy contains an empty entry")
 		}
 	}
 	for _, group := range []struct {
--- a/internal/sandboxrun/run.go
+++ b/internal/sandboxrun/run.go
@@ -3,7 +3,9 @@ package sandboxrun
 import (
 	"fmt"
 	"io"
+	"net/url"
 	"os"
+	"strings"
 
 	"github.com/tngtech/oh-my-agentic-coder/internal/audit"
 	"github.com/tngtech/oh-my-agentic-coder/internal/intent"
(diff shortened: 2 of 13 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,123 @@
+# CLAUDE.md
```
nhuelstng: pr #106 symlinked CLAUDE.md <-> AGENTS.md, würde die hier erstmal draußen lassen

nhuelstng:
review with fix suggestions in #113

nhuelstng (changes requested):

CLAUDE.md
GevatterGaul: removed

internal/sandboxprofile/profile.go
```diff
@@ -257,6 +272,17 @@ func (p *Profile) Validate() error {
 			return fmt.Errorf("sandbox profile: network.network_prompt.prompt_timeout_secs must be >= 0")
 		}
 	}
+	if p.Network.UpstreamProxy != "" {
+		u, err := url.Parse(p.Network.UpstreamProxy)
+		if err != nil || (u.Scheme != "http" && u.Scheme != "https") || u.Host == "" {
```
nhuelstng: **`https://` is accepted here but the dialer sends cleartext.**

Validation permits `https`, but `upstreamProxyDialer.DialTunnel` dials the proxy over plain TCP with no TLS wrapping (`internal/netproxy/dialer.go:148` — `nd.DialContext(ctx, "tcp", d.proxyURL.Host)`). Configuring `https://proxy:8080` therefore sends a cleartext `CONNECT` to a TLS listener → handshake garbage → every request 502s, with no hint the scheme is the cause.

Proven locally by feeding an `https://` URL to `NewUpstreamProxyDialer` and inspecting the first bytes on the wire: `"CONNECT example."` (first byte `C`, not `0x16` TLS ClientHello).

**Fix:** since TLS-to-proxy isn't implemented, reject `https` here (`want http://host`) — the smaller, honest change — or add `tls.Client` wrapping when `scheme == "https"`. `profile_test.go` currently asserts `https://proxy:8080` is *valid*; that case would flip.

internal/sandboxrun/run.go
```diff
@@ -217,6 +219,69 @@ func buildProxy(p *sandboxprofile.Profile, profilePath string, stderr io.Writer,
...
+	}
+	if proxyStr == "" {
+		proxyStr = os.Getenv("http_proxy")
+	}
+
+	if proxyStr == "" {
+		return netproxy.NewDirectDialer()
+	}
+
+	proxyURL, err := url.Parse(proxyStr)
```
nhuelstng: **Scheme-less env proxy silently kills all egress.**

`HTTPS_PROXY=host:port` (no scheme — the common corporate form Go's own `http.ProxyFromEnvironment` tolerates) parses without error: `url.Parse("proxy.corp:8080")` yields `Scheme="proxy.corp"`, `Host=""`. The resulting `upstreamProxyDialer` then routes every outbound connection to `""` and all traffic 502s. Note the profile-field path *is* guarded (`Validate` requires non-empty host), but this env path — the zero-config headline feature — is not.

Proven locally: `resolveUpstreamProxy` with `HTTPS_PROXY=proxy.corp:8080` returns a dialer whose `DialTunnel` fails with `upstream proxy : dial tcp: missing address` (empty `ProxyHost`).

**Fix:** after `url.Parse`, if `proxyURL.Host == ""`, prepend `http://` and re-parse (match stdlib), or warn and fall back to `NewDirectDialer()` like the invalid-URL branch already does.

nhuelstng (commented):
Two silent-failure config traps on the upstream-proxy entry points. Both reproduce empirically against this commit (local regression tests, red now / green after fix). Details inline.

nhuelstng (approved):

PR #116, opened 2026-07-17
fix(serve): allowlist the inner server's listen port in the sandbox (#115)

### LINKED ISSUE #115: omac serve: inner server's listen port is not allowlisted in the sandbox (opencode serve ServeError under tng-default)
## Summary

`omac serve <harness>` launches the inner **server daemon** (e.g. `opencode serve`) inside the sandbox, but never allowlists that server's **listen port** for `bind()`. Under any sandbox profile that doesn't already permit binding the port — notably the default `tng-default`, which pre-allows only `4097` and sets `network_prompt.on_unavailable: deny` — the inner server's `bind()` is denied and it crashes on startup. Desktop mode is unusable as a result.

## Repro

```
omac serve opencode            # under tng-default (or any profile with network_prompt: deny)
```

opencode aborts immediately after loading config:

```
Error: Unexpected error
ServeError
```

`omac serve opencode --no-sandbox` works fine, which localizes it to the sandbox.

## Root cause (strace)

`strace -f` of the sandboxed `opencode serve` shows the exact denial:

```
bind(17, {AF_INET, sin_port=htons(4096), 127.0.0.1}) = -1 EACCES (Permission denied)
bind(17, {AF_INET, sin_port=htons(0),    127.0.0.1}) = -1 EACCES (Permission denied)
```

`opencode serve` opens its HTTP listen socket on `127.0.0.1:4096` (then a random port as fallback). `omac serve` injects `--open-port` only for the **facade** and **control-plane** ports (for the sandboxed process to *connect* to omac); it never grants a `--listen-port`/`--open-port` for the inner server's *own* listen socket. In a headless daemon there is no UI to answer the `network_prompt`, so the bind is denied → `ServeError`.

Confirmed on omac 0.4.5 and current `main`; independent of the user's opencode config (reproduces with an empty config), harness version, and plugins/MCP.

## Fix (verified)

Allowlisting the port makes it work:

```
omac sandbox run --profile tng-default --listen-port 4096 -- opencode serve --port 4096   # ✅ listens
```

`omac serve` should inject the inner server's listen port into the sandbox argv (a `--listen-port`, mirroring the existing `--open-port` injection for the facade/control ports). Proposed: declare the port on the harness's `ServerLaunch` descriptor (`config.ServerLaunch`), so it stays data-driven and per-harness (opencode → `serve` on `4096`).

## Notes

- The weekly drift `serve` probe (#<serve-probe-PR>) uses a permissive profile, so it does not reproduce this; this is a product bug, not harness drift — it needs its own regression test (added with the fix).

nhuelstng:
Fixes #115.

## What

`omac serve <harness>` now allowlists the **inner server daemon's own listen port** in the sandbox, so its `bind()` is permitted. The port is declared per-harness on `config.ServerLaunch.ListenPort` (opencode's `serve` → `4096`); `serve.go` injects `--listen-port <port>` into the sandbox argv, mirroring the existing `--open-port` injection for the facade/control ports. A second commit adds a launch-time caution when that exposed port is unauthenticated.

## Why

`omac serve` opened `--open-port` only for the **facade** and **control-plane** ports (so the sandboxed process can *connect to* omac). The harness's server binds a **separate** listen socket that no grant covered. Under a restrictive profile — notably the default `tng-default` (permits listening only on `4097`, `network_prompt.on_unavailable: deny`) — the bind was denied non-interactively and `opencode serve` crashed on startup with an opaque `ServeError`, making Desktop mode unusable.

Root cause proven with `strace -f` of the sandboxed daemon:

```
bind(17, {AF_INET, sin_port=htons(4096), 127.0.0.1}) = -1 EACCES (Permission denied)
bind(17, {AF_INET, sin_port=htons(0),    127.0.0.1}) = -1 EACCES (Permission denied)
```

`--no-sandbox` worked; the crash reproduced with an empty opencode config, so it was neither the user's config/plugins/MCP nor an omac version regression.

## How

- `config.ServerLaunch` gains `ListenPort int` (`0` = none) and `AuthEnvVar string`. opencode declares `{ ListenPort: 4096, AuthEnvVar: "OPENCODE_SERVER_PASSWORD" }`; harnesses with no server mode declare nothing → no-op.
- `injectServerListenPort(argv, harness)` splices `--listen-port <port>` before the `--` separator, right after the control-port `--open-port` injection in `runServe`.
- `serverExposureWarning(harness, getenv)` emits a one-line caution before exec when the listen port is exposed and `AuthEnvVar` is unset.

## Security implications

The grant is deliberately minimal and does **not** weaken the sandbox:

- **Bind-only, loopback-only.** In `bwrap.go`, `ListenPort` maps to `--bind-tcp <port>` **only** — inbound bind permission for one loopback port. It grants **no** outbound connect (that's `--connect-tcp`, used by `OpenPort`), so it's *narrower* than the existing facade/control-port grants.
- **Core protections untouched.** Outbound domain filtering (the loopback proxy / `allow_domain`), filesystem confinement (bwrap binds + Landlock + protected paths), and secret isolation are all unchanged. Deny-by-default is preserved for every other bind/connect (`network_prompt` still denies them).
- **The one exposure:** opencode's server becomes reachable on `127.0.0.1:4096` (loopback only — **not** the LAN). It's unauthenticated unless `OPENCODE_SERVER_PASSWORD` is set, so a local process could drive the agent — but that agent is **still fully sandboxed** (fs + egress), so the blast radius is bounded, unlike `--no-sandbox`. This exposure is inherent to `opencode serve` (present whenever serve mode runs, with or without omac); the second commit surfaces it at launch and points at the mitigation.
- **Net:** strictly safer than the `--no-sandbox` workaround it replaces, which removes all isolation.

## Verification

- **TDD (red → green):** `TestServerLaunchListenPort` + `TestInjectServerListenPort` + `TestServerExposureWarning` written failing-first, then green. Full `internal/config` + `internal/cli` suites pass; `go vet` and `go build ./...` clean.
- **End-to-end:** a build from this branch runs `omac serve opencode` under the **real `tng-default`** profile without crashing (was `ServeError`); inner argv now carries `--listen-port 4096`; the exposure warning prints when `OPENCODE_SERVER_PASSWORD` is unset and is silent when set. Root cause and fix confirmed with `strace`.

## Notes / follow-ups

- Scope is intentionally minimal: it allowlists the harness's declared default port. A harness server on a non-default/random port would need that port to flow into the grant — out of scope here.
- The weekly drift `serve` probe (separate PR) uses a permissive profile, so it doesn't reproduce this — correct for a *harness-drift* probe; this is a product bug with its own unit coverage here.

### DIFF
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -405,6 +406,13 @@ func runServe(args []string, env *Env) int {
 		if cp := controlPortOf(cln); cp != "" {
 			argv = injectOpenPort(argv, cp)
 		}
+		// Allowlist the inner server daemon's own listen port so its bind()
+		// is permitted inside the sandbox. The profile's `--open-port
+		// {{tcp_port}}` covers only the facade; the harness's server (e.g.
+		// `opencode serve` on 4096) binds a distinct port that no other
+		// grant covers, so a restrictive profile (network_prompt: deny)
+		// denies the bind and the daemon crashes on startup (issue #115).
+		argv = injectServerListenPort(argv, harness)
 		// Grant the selected harness's runtime dirs (config, state,
 		// sessions) read+write — only for the selected harness.
 		argv = injectSandboxDirs(argv, harness.SandboxDirs)
@@ -495,6 +503,18 @@ func controlPortOf(ln net.Listener) string {
 		return ""
 	}
 	return port
+}
+
+// injectServerListenPort splices `--listen-port <port>` into a sandbox argv
+// for a harness whose server daemon binds a fixed loopback port (declared on
+// its ServerLaunch descriptor). Without this the daemon's bind() is denied
+// under a restrictive sandbox profile and it crashes on startup (issue #115).
+// Harnesses with no server mode, or none declaring a port, are a no-op.
+func injectServerListenPort(argv []string, h config.Harness) []string {
+	if h.ServerLaunch == nil || h.ServerLaunch.ListenPort <= 0 {
+		return argv
+	}
+	return injectSandboxFlag(argv, "--listen-port", strconv.Itoa(h.ServerLaunch.ListenPort))
 }
 
 // injectOpenPort splices `--open-port <port>` into a sandbox argv so the
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -458,6 +458,12 @@ func runServe(args []string, env *Env) int {
 		fmt.Fprintf(env.Stderr, "[verbose] inner argv: %v\n", argv)
 	}
 
+	// Caution when the harness server will expose an unauthenticated loopback
+	// port (fires whether sandboxed or not — the server listens either way).
+	if w := serverExposureWarning(harness, os.Getenv); w != "" {
+		fmt.Fprintln(env.Stderr, w)
+	}
+
 	// Run the inner command (opencode serve) in the sandbox, with the
 	// control plane already serving. ExecWithReady blocks until the child
 	// exits; the deferred facade/supervisor/control teardown then runs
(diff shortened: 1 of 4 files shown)


NoRiceToday (approved):
lgtm once those conflicts have been solved

PR #120, opened 2026-07-17
fix(netprompt): give network-access dialogs a readable size on all platforms

nhuelstng:
## What
Give the native network-access decision dialogs (`internal/netprompt/prompt.go`) a consistent, readable window size across platforms.

## Why
The zenity backend set only `--height 320` with **no `--width`**, so GTK auto-sized the window to its content and collapsed the radiolist below the width of its longest option (e.g. `Allow permanently (this host)`), truncating the choices — this is the "popups are too small on Ubuntu" symptom. The kdialog backend set **no size at all**.

## How
- Added shared `dialogWidth = 520` / `dialogHeight = 560` constants.
- zenity: added `--width` and drove `--height` from the constants.
- kdialog: added `--geometry 520x560`.
- macOS `choose from list` has no size API and already auto-sizes correctly, so it is left unchanged.

## Verification
- `go build ./...`
- `go vet ./internal/netprompt/`
- `go test ./internal/netprompt/`

### DIFF
--- a/internal/netprompt/prompt.go
+++ b/internal/netprompt/prompt.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -22,6 +23,17 @@ const (
 	tokenDenyPermanentHost    = "deny_permanent_host"
 	tokenDenyPermanentSuffix  = "deny_permanent_suffix"
 	tokenNeedsIntent          = "needs_intent"
+)
+
+// Dialog dimensions in pixels. The default GTK/Qt auto-size makes the popup
+// unreadable on Ubuntu: too narrow to show the longest option label ("Allow
+// permanently (this host)") and too short to show all seven options — the
+// multi-line prompt text consumes the height and the radiolist collapses to a
+// two-row scroll box. The width fits the labels; the height fits the prompt
+// text plus all seven rows plus the buttons without scrolling.
+const (
+	dialogWidth  = 520
+	dialogHeight = 560
 )
 
 // optionLabels are the exact seven dialog choices (nono parity, product
@@ -304,13 +316,17 @@ func (zenityBackend) available() bool {
 	return err == nil
 }
 
-func (zenityBackend) show(ctx context.Context, host string, port int, suffix, intent string) (string, error) {
+// zenityArgs builds the full zenity radiolist argv, including the explicit
+// window size. Extracted so the sizing can be asserted without launching a
+// dialog.
+func zenityArgs(host string, port int, suffix, intent string) []string {
 	args := []string{
 		"--list", "--radiolist",
 		"--title", "omac: network access",
 		"--text", promptText(host, port, intent),
 		"--column", "", "--column", "Decision",
-		"--height", "320",
+		"--width", strconv.Itoa(dialogWidth),
+		"--height", strconv.Itoa(dialogHeight),
 	}
 	for _, o := range optionLabels(suffix) {
 		sel := "FALSE"
@@ -319,7 +335,11 @@ func (zenityBackend) show(ctx context.Context, host string, port int, suffix, in
 		}
 		args = append(args, sel, o)
 	}
-	out, err := exec.CommandContext(ctx, "zenity", args...).Output()
+	return args
+}
+
+func (zenityBackend) show(ctx context.Context, host string, port int, suffix, intent string) (string, error) {
+	out, err := exec.CommandContext(ctx, "zenity", zenityArgs(host, port, suffix, intent)...).Output()
 	if err != nil {
 		// zenity exits 1 on Cancel; treat as cancel unless ctx expired.
 		if ctx.Err() != nil {
@@ -339,9 +359,13 @@ func (kdialogBackend) available() bool {
 	return err == nil
 }
 
-func (kdialogBackend) show(ctx context.Context, host string, port int, suffix, intent string) (string, error) {
+// kdialogArgs builds the full kdialog radiolist argv, including the explicit
+// --geometry window size. Extracted so the sizing can be asserted without
+// launching a dialog.
+func kdialogArgs(host string, port int, suffix, intent string) []string {
 	opts := optionLabels(suffix)
 	args := []string{
+		"--geometry", fmt.Sprintf("%dx%d", dialogWidth, dialogHeight),
 		"--title", "omac: network access",
 		"--radiolist", promptText(host, port, intent),
 	}
@@ -350,9 +374,14 @@ func (kdialogBackend) show(ctx context.Context, host string, port int, suffix, i
 		if o == "Deny once" {
 			state = "on"
 		}
-		args = append(args, fmt.Sprintf("%d", i), o, state)
+		args = append(args, strconv.Itoa(i), o, state)
 	}
-	out, err := exec.CommandContext(ctx, "kdialog", args...).Output()
+	return args
+}
+
+func (kdialogBackend) show(ctx context.Context, host string, port int, suffix, intent string) (string, error) {
+	opts := optionLabels(suffix)
+	out, err := exec.CommandContext(ctx, "kdialog", kdialogArgs(host, port, suffix, intent)...).Output()
 	if err != nil {
 		if ctx.Err() != nil {
 			return "", ctx.Err()
--- a/internal/netprompt/prompt.go
+++ b/internal/netprompt/prompt.go
@@ -25,16 +25,35 @@ const (
 	tokenNeedsIntent          = "needs_intent"
 )
 
-// Dialog dimensions in pixels. The default GTK/Qt auto-size makes the popup
+// Default dialog dimensions in pixels. The GTK/Qt auto-size makes the popup
 // unreadable on Ubuntu: too narrow to show the longest option label ("Allow
 // permanently (this host)") and too short to show all seven options — the
 // multi-line prompt text consumes the height and the radiolist collapses to a
 // two-row scroll box. The width fits the labels; the height fits the prompt
 // text plus all seven rows plus the buttons without scrolling.
+//
+// These are defaults: OMAC_PROMPT_WIDTH / OMAC_PROMPT_HEIGHT override them for
+// users on unusual displays (small/720p laptops where 560 px crowds the
+// viewport, tiling WMs, HiDPI) — see dialogDimensions.
 const (
 	dialogWidth  = 520
 	dialogHeight = 560
 )
+
+// dialogDimensions returns the popup width and height in pixels, honouring the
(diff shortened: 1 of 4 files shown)


nhuelstng:
## Update: added automated verification + fixed a height bug it caught

### What
Added two tiers of test coverage and bumped `dialogHeight` 360 → 560.

### Why
A vision-model e2e revealed the width fix alone was **insufficient**: the dialog was wide enough, but the multi-line prompt text consumed the height and the radiolist collapsed to a ~2-row scroll box — **5 of the 7 options were hidden** on Ubuntu. A pure argument-assertion test could never catch this; only rendering the real window does.

### How
- **Deterministic CI test** (`prompt_size_test.go`) — extracted `zenityArgs`/`kdialogArgs` and assert the popup carries `--width`/`--height` (zenity) and `--geometry` (kdialog) against sane minimums. Fast, no display, runs in normal CI.
- **Gated e2e** (`prompt_render_test.go`, `OMAC_RENDER_TEST=1`) — drives the real `show()` path, asserts the dialog **pops up** (an on-screen window appears), screenshots it, and has a vision model confirm **every** option label is fully visible/untruncated. The vision endpoint is provider-agnostic via `OMAC_VISION_BASE_URL` / `OMAC_VISION_MODEL` / `OMAC_VISION_API_KEY` (no provider specifics committed).
- Bumped `dialogHeight` to 560 so all seven options + prompt text + buttons render without scrolling.

### Verification
- `go test ./internal/netprompt/` — deterministic tests pass; e2e auto-skips without the gate.
- Ran the e2e locally against a vision model: dialog popped up, all 7 labels rendered, verdict `readable=true`.

nhuelstng:
## Update: ship tier 1 only + add a code-driven popup demo

### What
- Keep only the deterministic sizing test (`prompt_size_test.go`); removed the gated vision e2e.
- Added `cmd/netprompt-demo` + `scripts/netprompt-popup-demo.sh` to pop up the real dialog for visual inspection on any platform.

### Why
- Tier 1 is a plain `go test` — the existing CI already runs it via `go test ./...`, so it ships with **no workflow change** and **zero added CI cost**. The vision e2e's value didn't justify per-PR network/model coupling.
- A demo is the right tool for "see how it looks": the demo drives `netprompt.NewPrompter`/`Prompt` — the production code path — so any change to window size, labels, or prompt text is reflected automatically. No separate layout to keep in sync.

### How
```
scripts/netprompt-popup-demo.sh [-host H] [-port P] [-intent TEXT]
```
Falls back cleanly if no backend (zenity/kdialog/osascript) is installed.

### Verification
- `go build ./...`, `go vet ./...`, gofmt clean.
- `go test ./internal/netprompt/` passes; deterministic sizing tests run unconditionally in CI.
- Ran the demo on Ubuntu/Wayland: dialog shows the passed host/port/intent, derived suffix, and all 7 options fully visible at the new 520×560 size.

NoRiceToday:
Reviewing

internal/netprompt/prompt_size_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+)
+
+// Below these the GTK/Qt auto-size collapses the radiolist and truncates the
+// longest option label ("Allow permanently (this host)"). This is the
+// regression guard for the "popups too small on Ubuntu" fix.
+const (
+	minReadableWidth  = 480
+	minReadableHeight = 300
+)
```
NoRiceToday: **[MEDIUM] T1 — Regression-guard test floor is below the original buggy value**

`minReadableHeight = 300` is *below* the original buggy `--height 320`, so a future drift back to 310 — which the PR's own comment says collapses the radiolist to "a two-row scroll box" — would still pass `TestDialogDimensionsAreReadable`. The test guards a magic-number floor, not the value the PR's rationale actually requires.

Suggest either raising the floors to the actual derived minimums and documenting *why* (see T2), or asserting equality with the constants and deriving the floor from the longest label width so the test fails if labels grow without the width growing.

internal/netprompt/prompt_size_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+func TestDialogDimensionsAreReadable(t *testing.T) {
+	if dialogWidth < minReadableWidth {
+		t.Errorf("dialogWidth = %d, want >= %d", dialogWidth, minReadableWidth)
+	}
+	if dialogHeight < minReadableHeight {
+		t.Errorf("dialogHeight = %d, want >= %d", dialogHeight, minReadableHeight)
+	}
+}
```
NoRiceToday, line 67: **[MEDIUM] T2 — No test pins `dialogWidth` ≥ longest label width (the actual bug)**

The bug being fixed is "too narrow to show the longest option label." The suite never measures any label. `optionLabels("example.com")` produces `"Allow permanently (*.example.com)"` (32 chars) and there is no assertion that `dialogWidth` is large enough to render it. If a future PR lengthens the labels (e.g. i18n, or `"Allow permanently (host: api.github.com)"`) without bumping `dialogWidth`, the truncation bug silently returns while `TestDialogDimensionsAreReadable` stays green at 520.

The test guards a proxy (a magic number ≥ another magic number), not the property (label fits). The property is the thing that broke.

Suggest adding a test that estimates rendered label width (character count × a conservative per-char factor, e.g. 9 px for the default GTK/Qt UI font) and asserts `dialogWidth` exceeds the longest label's estimated rendered width plus paddings/columns. Even a rough `dialogWidth > len(longestLabel)*8 + 120` heuristic would catch the real regression class.

internal/netprompt/prompt.go
```diff
@@ -24,6 +25,17 @@ const (
...
+// Dialog dimensions in pixels. The default GTK/Qt auto-size makes the popup
+// unreadable on Ubuntu: too narrow to show the longest option label ("Allow
+// permanently (this host)") and too short to show all seven options — the
+// multi-line prompt text consumes the height and the radiolist collapses to a
+// two-row scroll box. The width fits the labels; the height fits the prompt
+// text plus all seven rows plus the buttons without scrolling.
+const (
+	dialogWidth  = 520
+	dialogHeight = 560
+)
```
NoRiceToday, line 41: **[MEDIUM] C1 — Fixed-pixel sizing can overflow small / HiDPI displays (no override)**

`dialogHeight = 560` px on a 1366×768 laptop with a typical ~24 px title bar + ~40 px taskbar leaves ~704 px usable; 560 fits but leaves a thin margin. On a 720p (1280×720) display the dialog + chrome approaches the full viewport, and on a tiling WM (i3/Sway) that auto-centers, a 560-px-tall dialog can overlap or push other windows. On HiDPI/4K the *logical* pixel is what GTK/Qt use, so 520×560 looks small — fine — but the PR gives no way for a user on an unusual display to override. There is no env var (`OMAC_PROMPT_WIDTH`/`OMAC_PROMPT_HEIGHT`) and no config field.

The PR trades "too small" for "potentially too tall on small screens" and removes no configurability that didn't exist before, but it adds a new hardcoded value where the codebase otherwise reads sizing from config (`netprompt.LoadLearnedPolicy`, `sandboxprofile`). For an accessibility-sensitive product this is a gap.

Suggest either (a) clamping `dialogHeight` to a fraction of the viewport (requires a display query — out of scope for this PR), or (b) accepting env overrides `OMAC_PROMPT_WIDTH`/`OMAC_PROMPT_HEIGHT` parsed in `NewPrompter` with the constants as defaults. (b) is ~6 lines and closes the accessibility hole. Not a blocker for merge; a follow-up is acceptable.

internal/netprompt/prompt.go
```diff
@@ -350,9 +374,14 @@ func (kdialogBackend) show(ctx context.Context, host string, port int, suffix, i
 		if o == "Deny once" {
 			state = "on"
 		}
-		args = append(args, fmt.Sprintf("%d", i), o, state)
+		args = append(args, strconv.Itoa(i), o, state)
 	}
-	out, err := exec.CommandContext(ctx, "kdialog", args...).Output()
+	return args
+}
```
NoRiceToday, line 409: **[MEDIUM] C2 — kdialog `--geometry` semantics & ordering unverified**

The args are built as `["--geometry", "520x560", "--title", "...", "--radiolist", "<promptText>..."]`. Two concerns I could not verify from the sandbox (no kdialog binary, manpage fetches 404'd):

1. **kdialog `--geometry` format.** kdialog (KDE) historically uses the X11 geometry spec `WIDTHxHEIGHT+X+Y`, e.g. `--geometry 520x560`. The `WxH` form *should* work, but some kdialog versions parse `--geometry` as `=<value>` (i.e. `--geometry=520x560`). If the installed kdialog is the KDE6/Qt6 build, behaviour may differ from KDE5. **Unverified.**

2. **Argument ordering.** kdialog requires the dialog-type flag (`--radiolist`, `--checklist`, etc.) and its immediate positional argument (the prompt text) to come *before* the option triples. Placing `--geometry` *before* `--radiolist` is the conventional position for global options, but if kdialog parses `--geometry` greedily and consumes the next token as part of the geometry string, the ordering is fine; if it expects `--geometry` *after* the dialog subcommand, this could silently mis-parse. **Unverified.**

If either is wrong, kdialog either ignores the size (bug not fixed on KDE) or errors on launch (worse: every KDE prompt fails → deny-once → network breaks).

Suggest the author run `scripts/netprompt-popup-demo.sh` on a real KDE session (or at least `kdialog --geometry 520x560 --radiolist "test" 0 "a" off 1 "b" on`) and confirm the window is the requested size. The CI matrix (`.github/workflows/e2e.yml`) runs Linux but the e2e suite uses the stub backend, so it does *not* exercise the real kdialog path — this is a genuine coverage gap. At minimum, add a comment citing the kdialog version tested.

NoRiceToday:
**macOS: consider a larger and prettier window too**

`internal/netprompt/prompt.go:184-186` (osascript backend, unchanged by this PR)

This PR leaves the macOS `choose from list` path alone, which is correct in the sense that AppleScript exposes no size API for it. But the macOS dialog is the one place we still ship the auto-sized, default-styled list — small buttons, small system font, no product identity. Even without a real size API there are levers:

- Wrap `choose from list` in `display dialog` with `with title` / `with prompt` and a custom OK button, which gives more control over button size and dialog chrome.
- Or move the macOS path off `choose from list` entirely to an AppleScript `choose from list` inside a `display dialog` with `buttons` / `default button` / `cancel button` / `with icon`, which gives a larger, better-looking window that still auto-sizes to content.
- If we want a real window size on macOS, the clean answer is a tiny SwiftUI/Security-Scoped Swift wrapper invoked via `osascript -l JavaScript` or a bundled helper — but that is a larger follow-up, not this PR.

Not a blocker — flagging because the PR title says "all platforms" and macOS is the one platform still on the default tiny system dialog. A follow-up issue for "macOS dialog: larger and prettier" would close the gap.

PR #125, opened 2026-07-18
Fix/serve xdg state home

thomas-endres-tng:
Running omac serve --for-opencode-desktop from a plain shell (rather than being launched by the OpenCode Desktop app) left XDG_STATE_HOME unset, so the sandboxed opencode read an empty state directory, its providers resolved without credentials, and /config/providers returned a 500 — which surfaces to clients like opencode attach as an opaque "Unexpected server error". This defaults XDG_STATE_HOME to the Desktop app's data directory when it isn't already set (an explicit value is respected), so omac's own provisioning and the inner opencode both resolve state against the same store, and it grants that directory read+write in the sandbox. The directory is resolved per-OS from the Electron userData convention (macOS Application Support, Linux ~/.config / XDG_CONFIG_HOME), preferring whichever install actually exists on disk.

It also parallelizes the filesystem.deny scan that runs during sandbox setup. That scan walks every granted root to mask secret basenames such as .env and .git, and under --for-opencode-desktop — dozens of project folders plus large cache and toolchain directories — it was walking them one at a time, which dominated launch. Walking the roots concurrently (bounded by CPU count) is behavior-preserving, since the resulting match set is deduplicated and sorted by callers, and it cut startup from roughly 24s to 11s in local testing. Unit tests, including the race detector, pass.

### DIFF
--- a/internal/cli/serve.go
+++ b/internal/cli/serve.go
@@ -147,6 +147,29 @@ func runServe(args []string, env *Env) int {
 		return ExitMisuse
 	}
 
+	// --for-opencode-desktop: the OpenCode Desktop app runs opencode with
+	// XDG_STATE_HOME pointing at its own data dir (auth, credentials,
+	// config), and both omac's desktop provisioning AND the inner
+	// opencode must resolve state against that same store. When the
+	// Desktop launches omac it exports XDG_STATE_HOME for us; a manual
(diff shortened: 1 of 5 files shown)


nhuelstng:
lgtm; nur noch dco fixen

PR #126, opened 2026-07-18
test(sandbox): e2e that external tools reach network via the omac proxy

nhuelstng:
## What

Adds end-to-end coverage that a **real external tool**, run inside a filtered bwrap sandbox, reaches the network **only through the omac filtering proxy** and **only for allowlisted hosts**. Stacks on top of #122.

Two tests (`internal/sandboxrun/proxyinject_integration_linux_test.go`):
- **`TestIntegrationCurlThroughOmacProxy`** — curl (proxy-aware, honors `HTTP(S)_PROXY`): allowlisted host tunnels through; non-allowlisted host is denied by the filter.
- **`TestIntegrationNodeFetchThroughOmacProxy`** — the `node` `proxy_injection` family: Node's native `fetch` (undici) ignores `HTTP(S)_PROXY` until `NODE_USE_ENV_PROXY=1` — the exact var the injector sets — so this proves the *mechanism*, not just the env string. Requires Node ≥ 24.

Plus a dedicated **"External tooling"** CI job in `ci.yml` (bubblewrap + Landlock ABI ≥ 4 + curl + Node 24).

## Why

#122 had three test layers but a gap between them: in-process proxy tests (`internal/netproxy`) and env-string unit tests (`proxyinject_test.go`), but **nothing driving a real tool through the omac CONNECT proxy inside a real sandbox**. That end-to-end path — Landlock enforcement + proxy env injection + a real toolchain — was unverified. These tests close it and act as a regression guard for "common tools remain usable under a filtered sandbox".

The node case specifically verifies that `NODE_USE_ENV_PROXY` actually routes native `fetch` through the proxy at runtime — the unit test only asserts the string is produced.

## How

- **Hermetic, cheap, no network / no LLM / no token**: a loopback `httptest` upstream, and a proxy filter that resolves a fake host (`repo.omac-e2e.test`) to `127.0.0.1` (loopback-CONNECT refusal keys on the requested *hostname*, not the resolved IP).
- **The load-bearing invariant**: Landlock allows only the proxy's loopback port (`Stage2Args` emits `--connect-tcp <ProxyPort>`); the upstream's port is *not* allowlisted, so a direct dial is kernel-blocked. Therefore a `200` can **only** arrive via the proxy — the positive assertion genuinely proves proxy routing, not incidental reachability.
- **Isolated from the main test job**: gated behind the `external_tools` build tag and run in a separate CI job, so `go test ./...` stays free of the curl/Node-24 runtime dependency. Every precondition — missing curl, missing node, Node < 24, **and Landlock ABI < 4** — goes through `skipOrFailCI`: skips locally, **fails** in CI so a misconfigured runner can't silently drop coverage.
- **Assertions pin to the filter's verdict, not the tool's exit status.** The proxy's `FilterConfig.Logf` is captured, and each case asserts the recorded `ALLOW`/`DENY` for that host. A tool that fails for an unrelated reason (sandbox never launched, missing read grant) leaves no `DENY` behind and so cannot pass as a working denial. This matters most for node: undici reports a denied `CONNECT` generically (`fetch failed` / `Request was cancelled`) with no status to match on.
- **The child env is built the way `Run` builds it** — `sandboxprofile.FilterEnv(environ, nil, injected)` — rather than handing bwrap a raw `os.Environ()`. Deliberate ambient poison is prepended (a blocklisted `JAVA_TOOL_OPTIONS`, plus `NODE_USE_ENV_PROXY=0` and `HTTP(S)_PROXY` pointing at TEST-NET-1), so the blocklist-then-overlay ordering that makes injection work is under test instead of assumed.

## Verification

Rebased onto the current `feat/network-proxy-injection` tip and re-run there
(bwrap 0.11 + Landlock ABI ≥ 4 + curl 8.18 + Node 24.14):

```
$ go test -tags=external_tools -run 'ThroughOmacProxy' -v -count=1 ./internal/sandboxrun/
--- PASS: TestIntegrationCurlThroughOmacProxy (2.11s)
--- PASS: TestIntegrationNodeFetchThroughOmacProxy (1.68s)
PASS
```

**The load-bearing invariant was verified directly, not just asserted.** Same
sandbox, same upstream, no proxy env in the child: `curl: (7) Failed to
connect to 127.0.0.1 port 43747`. Only the proxy port is Landlock-allowed, so
the `200` in the positive case can only have arrived through the proxy.

**The new assertions were mutation-tested** — both mutants fail, confirming
they catch real regressions:

| Mutation | Result |
|---|---|
| drop the injected env overlay (ambient poison wins) | **FAIL** — verdict `""` for both hosts; the *previous* assertion passed this on the denied-host case |
| filter allows the denied host | **FAIL** — verdict `ALLOW`, want `DENY` |

- `go vet -tags=external_tools ./internal/sandboxrun/` clean; `gofmt` clean.
- `go test ./internal/sandboxrun/` (no tag) reports **no tests to run** for these — confirms the tag keeps them off the main gate.
- `ci.yml` parses as valid YAML; jobs are `test, wsl2-session, lint, e2e-fast, cache-integration, external-tooling`. The rebase conflict in `ci.yml` was purely additive (both branches append a job) — `cache-integration` is preserved.
- The new job also runs `go vet -tags=external_tools`: the `lint` job runs staticcheck without build tags, so these files were otherwise unanalysed.

> [!IMPORTANT]
> **Base branch.** This targets `feat/network-proxy-injection`, and `ci.yml`
> only triggers on `pull_request: branches: [main]` — so CI does not run here
> and only DCO reports. Retarget to `main` once #122 merges to get the new
> job actually exercised; retargeting earlier would fold #122's diff into
> this PR and break the stacked review.

Refs #122

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,13 +207,15 @@ jobs:
   external-tooling:
     # Proves proxy-unaware / proxy-aware external toolchains actually reach
     # the network through the omac filtering proxy under a filtered bwrap
-    # sandbox (build tag `external_tools`): curl (honors HTTP(S)_PROXY) and
+    # sandbox (build tag `external_tools`): curl (honors HTTP(S)_PROXY),
     # Node's native fetch via the `node` proxy_injection family
-    # (NODE_USE_ENV_PROXY, Node >= 24). Split out from the main test job
-    # because it needs a real tool matrix (bubblewrap + Landlock ABI >= 4 +
-    # curl + Node 24) that the hermetic unit tests don't; keeping it separate
-    # keeps `go test ./...` free of that runtime dependency. Linux-only: the
-    # kernel proxy/Landlock path is Linux-specific.
+    # (NODE_USE_ENV_PROXY, Node >= 24), and the JVM via the `jvm`
+    # proxy_injection family (JAVA_TOOL_OPTIONS, any modern JDK). Split out
+    # from the main test job because it needs a real tool matrix (bubblewrap
+    # + Landlock ABI >= 4 + curl + Node 24 + JDK) that the hermetic unit
+    # tests don't; keeping it separate keeps `go test ./...` free of that
+    # runtime dependency. Linux-only: the kernel proxy/Landlock path is
+    # Linux-specific.
     name: "External tooling (proxy egress)"
     runs-on: ubuntu-latest
     steps:
@@ -244,6 +246,15 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 24
+
+      # The jvm proxy_injection family test compiles a small Java probe with
+      # javac and runs it with java inside the sandbox. Temurin is the default
+      # JDK on ubuntu-latest via setup-java; javac is bundled in the JDK.
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '21'
 
       # The lint job runs staticcheck without build tags, so it never sees
       # the external_tools files. Vet them here or they go unanalysed.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,3 +203,52 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: go test -count=1 -run '^TestIntegration.*Cache' ./internal/sandboxrun
       - run: go test -tags=e2e -count=1 -timeout=15m -run '^TestE2ECache' ./internal/e2e/
+
+  external-tooling:
+    # Proves proxy-unaware / proxy-aware external toolchains actually reach
+    # the network through the omac filtering proxy under a filtered bwrap
+    # sandbox (build tag `external_tools`): curl (honors HTTP(S)_PROXY) and
+    # Node's native fetch via the `node` proxy_injection family
+    # (NODE_USE_ENV_PROXY, Node >= 24). Split out from the main test job
+    # because it needs a real tool matrix (bubblewrap + Landlock ABI >= 4 +
+    # curl + Node 24) that the hermetic unit tests don't; keeping it separate
+    # keeps `go test ./...` free of that runtime dependency. Linux-only: the
+    # kernel proxy/Landlock path is Linux-specific.
+    name: "External tooling (proxy egress)"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install bubblewrap
+        run: |
+          sudo apt-get update && sudo apt-get install -y bubblewrap
+          # Ubuntu 24.04 AppArmor restricts unprivileged user namespaces;
+          # grant bwrap the userns permission (see the test job above) so the
+          # sandbox launches instead of silently skipping via requireBwrap.
+          sudo tee /etc/apparmor.d/bwrap > /dev/null <<'EOF'
+          abi <abi/4.0>,
+          /usr/bin/bwrap flags=(unconfined) {
+            userns,
+          }
+          EOF
+          sudo apparmor_parser -r /etc/apparmor.d/bwrap
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+          cache: true
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 24
+
+      # The lint job runs staticcheck without build tags, so it never sees
+      # the external_tools files. Vet them here or they go unanalysed.
+      - name: Vet tagged sources
+        run: go vet -tags=external_tools ./internal/sandboxrun/
+
+      - name: Run external-tooling proxy tests
+        run: go test -tags=external_tools -count=1 -timeout=10m ./internal/sandboxrun/
--- a/internal/sandboxrun/proxyinject_integration_linux_test.go
+++ b/internal/sandboxrun/proxyinject_integration_linux_test.go
@@ -1,0 +1,251 @@
+//go:build linux && external_tools
+
+package sandboxrun
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/netip"
+	"net/url"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/tngtech/oh-my-agentic-coder/internal/netproxy"
+	"github.com/tngtech/oh-my-agentic-coder/internal/sandboxprofile"
+)
+
+// These integration tests prove that a *real tool* run inside a filtered
+// bwrap sandbox reaches the network only through the omac proxy, and only
+// for allowlisted hosts. They are the end-to-end counterpart to the
+// in-process proxy tests (internal/netproxy) and the env-string unit tests
+// (proxyinject_test.go): here curl and node actually CONNECT through the
+// proxy under Landlock enforcement.
+//
+// Hermetic: the upstream is a loopback httptest server and the proxy filter
+// resolves a fake hostname to 127.0.0.1 (the loopback-CONNECT refusal keys
+// on the requested hostname, not the resolved IP — see
+// netproxy/server_test.go). No real network, no LLM, no token; runs on the
+// PR `go test ./...` gate.
+
+const (
+	proxyTestAllowedHost = "repo.omac-e2e.test"
+	proxyTestDeniedHost  = "blocked.omac-e2e.test"
+)
+
+// startHermeticProxy builds an omac filtering proxy that allows exactly
+// proxyTestAllowedHost and resolves every hostname to 127.0.0.1, so a
+// loopback httptest upstream stands in for the allowlisted repository.
+func startHermeticProxy(t *testing.T) *netproxy.Server {
+	t.Helper()
+	filter := netproxy.NewFilter(netproxy.FilterConfig{
+		AllowDomains: []string{proxyTestAllowedHost},
+		Resolve: func(_ context.Context, _ string) ([]netip.Addr, error) {
+			return []netip.Addr{netip.MustParseAddr("127.0.0.1")}, nil
+		},
+	})
+	srv, err := netproxy.NewServer(filter, netproxy.NewDirectDialer(), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := srv.Start(); err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(srv.Close)
+	return srv
+}
+
+// buildOmacBinary compiles the real omac binary; stage2 must be a genuine
+// `omac sandbox stage2` invocation (the test binary cannot stand in).
+func buildOmacBinary(t *testing.T) string {
+	t.Helper()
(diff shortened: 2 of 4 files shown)


nhuelstng:
## Review follow-ups applied

Rebased and pushed as `f78c6d9`. The branch now shows **only its own 2 commits** and is `MERGEABLE` (was `CONFLICTING`).

### 1. The rebase was the blocker — the branch was reverting live features

`test/e2e-proxy-injection-tools` branched from `694cfd0` (#121) while #122 branched from `a77b82e` (#106) — ~50 commits of drift, and #122 had also been force-pushed since. Diffing `run.go` between the tips, this branch's copy was missing:

- the entire tool-cache redirect (`injectedToolCacheEnv`, `canonicalCachePath`, `OMAC_CACHE_DIR`/`_MODE`) — #104/#158
- `harnessName()` + `hostmap.For(harness)` + `origin.NewResolver()` netprompt attribution — #132
- the `resolveUpstreamProxy` scheme-less-URL fix, reverted to the unconditional `"http://" + proxyStr` — #124
- all of #122's own review fixes (the dead `https.nonProxyHosts` sysprop, the Node<24 warning, the `ExecWithEnv` workdir arg)

That last one showed up empirically: running a JVM probe in this tree emitted `-Dhttps.nonProxyHosts=…`, the exact sysprop #122 removed as dead. Post-rebase all six are confirmed present, `go test ./...` is clean, and the tag still keeps these tests off the default gate (`no tests to run` untagged).

### 2. Assertions now pin to the filter's verdict

The negative checks were `code == 0 && strings.Contains(out, "200")`, which passes on *any* failure — a sandbox that never launched read as a working denial. Both tests now capture the proxy's `FilterConfig.Logf` and assert the recorded `ALLOW`/`DENY` per host.

This matters most for node: undici reports a denied `CONNECT` as `fetch failed` / `Request was cancelled` with **no status to match on**, so there is no string to assert. The filter's own `DENY` is the only precise signal. (curl does surface the `403`, and that's asserted too.)

Mutation-tested to confirm the assertions bite:

| Mutation | Result |
|---|---|
| drop the injected env overlay | **FAIL** — verdict `""` for both hosts. The *old* assertion passed this on the denied-host case. |
| filter allows the denied host | **FAIL** — verdict `ALLOW`, want `DENY` |

### 3. The child env is now built the way `Run` builds it

Was `cmd.Env = append(os.Environ(), extraEnv...)`, which proved the mechanism but skipped omac's real env pipeline — where the interesting regression risk lives. Now `sandboxprofile.FilterEnv(environ, nil, injected)`, with deliberate ambient poison prepended: a blocklisted `JAVA_TOOL_OPTIONS`, plus `NODE_USE_ENV_PROXY=0` and `HTTP(S)_PROXY` at TEST-NET-1 (`192.0.2.1:9`). Both tests pass *with* the poison, so blocklist-then-overlay ordering is under test rather than assumed — and it also removes the flake risk from a dev machine's ambient `NO_PROXY`.

### 4. Gates now fail closed

Missing `curl` was a plain `t.Skip` and Landlock ABI < 4 was `t.Skipf` — only the node paths used `skipOrFailCI`, despite the PR body claiming otherwise. A runner image without curl or with an older kernel would have turned this job green while testing nothing. All four preconditions now route through `skipOrFailCI`.

### 5. Smaller items

- Dropped the file-header claim that these "run on the PR `go test ./...` gate" — the build tag exists precisely so they don't. Replaced with what actually gates them, plus why the positive assertions are load-bearing.
- The `ci.yml` rebase conflict was purely additive (both branches append a job); `cache-integration` is preserved alongside `external-tooling`.
- Added `go vet -tags=external_tools` to the new job: `lint` runs staticcheck without build tags, so these files were unanalysed. (`gofmt -l .` did cover them.)

### The central claim holds — verified directly

Probed the load-bearing invariant rather than trusting it: same sandbox, same upstream, no proxy env in the child → `curl: (7) Failed to connect to 127.0.0.1 port 43747`. Only the proxy port is Landlock-allowed, so the `200` in the positive case can only have come through the proxy. The positive assertions are genuinely non-vacuous.

### Still open: base branch

`ci.yml` triggers only on `pull_request: branches: [main]`, so **CI still does not run here** — only DCO reports. The new job remains CI-unexercised (it passes locally with the matrix it declares). Retarget to `main` once #122 merges; doing it now would fold #122's diff into this PR and break the stacked review. Flagged in the body as well.

NoRiceToday:
## Adversarial review follow-ups

Based on the in-depth adversarial review, addressed the following findings (commits `f20b1af` and `e8cf18d`):

### P1 — `verdictFor` could conflate a policy DENY with an unrelated DENY
`verdictFor` matched on the substring `"net DENY"` + `host":"`, so any DENY line for the host — including an unrelated auth-failure or hard-deny — would satisfy the assertion. Changed `verdictFor` to return `(verdict, reason)` via a new `extractReason` helper, and pinned the denial assertions in both tests to `reason == "not in allowlist"` — the only verdict this filter config can legitimately produce for the denied host. This closes the gap so a future wiring change or new DENY reason can't silently satisfy the assertion, matching the test's own stated philosophy.

*Correction on the original framing:* the reviewer initially claimed an auth-failure DENY (`server.go:214`) could satisfy the assertion today. That was wrong — the test passes `nil` for the server-level `logf`, so auth-failure lines never reach `dec.lines`. The fix is honest hardening, not a live-bug fix.

### P2 — Blocklist exercised but never asserted
`ambientPoison` includes `JAVA_TOOL_OPTIONS` (blocklisted) and poisoned proxy vars (overridden by `injected`), but nothing verified the child env actually lacked/overrode them. A regression in `dangerousEnvExact` would pass the tool-based tests silently. Added `TestEnvLayeringDropsBlocklistAndOverlays`: a bwrap-free pure-env test that asserts `JAVA_TOOL_OPTIONS` is absent and `HTTP_PROXY` equals the injected value.

### P2 — JVM `proxy_injection` family had no integration test
The `jvm` family — the original motivation for the feature, with the most complex injector — had only env-string unit tests. Added `TestIntegrationJVMThroughOmacProxy`: precompiles a minimal Java HTTPS client, runs `java` inside the sandbox with the `jvm` family injected, asserts the filter records ALLOW for the allowlisted host and DENY (`not in allowlist`) for the denied host. The bare JDK ignores `proxyUser`/`proxyPassword` (documented in `proxyinject.go:JVMProxyToolOptions`), so the allowlisted CONNECT is filter-ALLOWed but the proxy returns 407 — the filter's verdict is the pin, not an HTTP 200. CI: added a Temurin JDK 21 setup step to the `external-tooling` job.

### Verification
- `gofmt` clean; `GOOS=linux go test -tags=external_tools` compiles clean.
- Non-tagged `go test ./internal/sandboxrun/` reports "no tests to run" (tag still keeps them off the main gate).
- `ci.yml` parses as valid YAML; all 6 jobs present.

nhuelstng:
## CI failure was a real product bug, not a test bug

The `External tooling (proxy egress)` job was failing:

```
--- FAIL: TestIntegrationJVMThroughOmacProxy
    java: error while loading shared libraries: libjli.so:
    cannot open shared object file: No such file or directory
```

Verified empirically before touching anything, and it is not confined to the test.

### 1. `resolveInnerBinaryDirs` under-grants prefix-layout tools

Reproduced against **production** code, not the test's copy: copied a JDK to a
non-baseline path and ran it through the real grant resolution —

```
resolveInnerBinaryDirs("~/omac-jdk-repro/bin/java") = [~/omac-jdk-repro/bin]
exit=127  java: error while loading shared libraries: libjli.so
```

Only the binary's own directory is granted, so a tool whose runtime lives
*beside* its bin dir cannot start. Bisected the required set by granting
sibling dirs one at a time:

| Granted | Result |
|---|---|
| `bin` | `error while loading shared libraries: libjli.so` |
| `bin`+`lib` | starts; `java -version` OK, but TLS dies: `InternalError: Error loading java.security file` |
| `bin`+`lib`+`conf` | fully working |

**Why CI caught it and local runs did not:** the gap is invisible whenever the
install root falls under a baseline read grant. A distro JDK lives in
`/usr/lib/jvm/...` and `/usr` is baseline, so it works. The runner's
`setup-java` JDK is at `/opt/hostedtoolcache/...` and is not. This equally
breaks a JDK under `~/.sdkman`, `~/.jdks`, or a mise/asdf install, and Python
venvs have the same shape — so this is a live sandbox bug for real users, not
CI-only.

**Fix (TDD, red first):** derive the install prefix from a conventional
`bin`/`sbin`/`libexec` dir and grant its existing `lib`/`lib64`/`libexec`/`conf`
subdirs. The prefix root itself is never granted, so sibling trees (`share`,
`src`, a checkout) stay invisible. `$HOME` and its ancestors are never treated
as a prefix — a personal `~/bin` must not widen the grant to `~/lib`. Those two
guards are tests, not just comments.

Confirmed with the exact CI configuration locally (Temurin 21 at a non-baseline
path): the failure reproduces byte-for-byte without the fix, and passes with it.
Also re-ran a real harness launch through `omac start opencode -- --version`
(exit 0) to confirm the wider grants do not regress the normal path.

### 2. The JVM test's assertions were unreachable by construction

With `java` now starting, the test still failed — and revealed the deeper
problem: **zero filter decisions were ever recorded** (`decisions: []`) while
the output showed `STATUS 407`.

Cause: `netproxy.Server.handle` checks proxy auth *before* dispatching CONNECT
(`server.go:213`). The bare JDK ignores the `proxyUser`/`proxyPassword` system
properties, so every request — allowlisted or not — was answered 407 at the
auth stage and the filter never ran. The test asserted `ALLOW` for one host and
`DENY` for the other; neither could ever be produced. The denied-host case also
exited 0, since the probe printed `STATUS 407` and returned normally.

This is the same area as the earlier P1 note, but the conclusion there was
inverted: the auth-failure path was not merely unable to *satisfy* the
assertion, it made the assertion unsatisfiable.

**Fix:** the probe now installs the `Authenticator` that Gradle and Maven
install from those properties — the documented real consumers of the `jvm`
family — plus a trust-all manager for the loopback httptest cert (the same
concession `curl -k` and `NODE_TLS_REJECT_UNAUTHORIZED=0` already make). The
test now asserts a real `STATUS 200` through the proxy and `DENY (not in
allowlist)` for the blocked host, making it as strong as the curl and node cases.

### 3. Removed the duplication that allowed the drift

The helpers hand-rolled their own grant logic, which is precisely why they
could diverge from production. They now call `resolveInnerBinaryDirs`, so the
tests exercise production's answer to "what does this binary need". The
near-duplicate `runToolWithExtraReads` is folded into `runToolThroughProxy`.

### Verification

Mutation-tested the JVM assertions — both mutants fail:

| Mutation | Result |
|---|---|
| revert the production grant fix | **FAIL** — `libjli.so`, verdict `""` |
| filter allows the denied host | **FAIL** — verdict `ALLOW`, want `DENY` |
| probe installs no `Authenticator` | **FAIL** — verdict `""` (the 407 path) |

- `go test ./...` clean; `go vet ./...` and `go vet -tags=external_tools ./...` clean; `gofmt -l .` empty.
- All four tagged tests pass locally with the CI matrix (bwrap, Landlock ABI ≥ 4, curl, Node 24, Temurin 21 at a non-baseline path): `TestIntegrationCurlThroughOmacProxy`, `TestIntegrationNodeFetchThroughOmacProxy`, `TestIntegrationJVMThroughOmacProxy`, `TestEnvLayeringDropsBlocklistAndOverlays`.
- No `ci.yml` change needed — the fix covers the `hostedtoolcache` path the job already installs to.

PR #128, opened 2026-07-18
feat(diagnose): add `omac diagnose` — explain why a sandboxed run failed

nhuelstng:
## What

Adds **`omac diagnose`**, the *reactive* counterpart to `omac doctor` (doctor = "is my setup correct?"; diagnose = "why did my run fail?"). It reads the previously **write-only** audit trail back and correlates the sandbox's actual network decisions against the effective policy. It is **pure, read-only inspection** — it never launches a sandbox or contacts the network.

- **Surfaces what the sandbox blocked** — groups `net.decision` denials from `audit.jsonl` (structured, persistent, CI-safe), scoped to the last run.
- **Explains non-obvious clashes** — a host in `allow_domain` yet DENIED (names the shadowing deny/learned/hard-deny rule); unused/typo `allow_domain` entries; DNS failures.
- **Names the invisible blind spots** — proxy-unaware tools (JVM/gradle connecting directly), raw-TCP/SSH (`allow_tcp_connect`), and filesystem/socket/loopback denials (docker.sock, local services) that leave no trace in the network log.
- **`--probe host[:port]`** — static admission check reusing the real filter (no DNS, no dialing, no sandbox launch): reports the HTTP(S)-via-proxy path *and* the raw-TCP path; loopback-aware. The verdict cannot drift from runtime behavior because it evaluates the same `netproxy` filter the sandbox enforces.
- **Focused output** — one-line status → problems (with fixes) → collapsed advisories; `-v` expands everything; `--json` for a shareable bundle.
- **`omac doctor`** gains the existing `profileaudit` security lint (advisory) and points at `diagnose`.

## Why

Every network decision omac makes is already recorded, but no command read it back — so when gradle/npm/kubectl/etc. failed inside the sandbox, users had no way to see *what* was blocked or *which rule* did it. The doctor can't cover every external tool; the design surfaces **evidence + generic, least-privilege hints** so a user can debug an unknown tool themselves. It reuses `provenance` (effective config) and `profileaudit` (security lint) rather than reimplementing them.

**Least-privilege by design:** hints recommend the *exact* host/port, legitimize leaving a host denied, and flag over-broad/unused rules for removal — never "allow everything".

## How

- `internal/diagnose` — pure, stdlib-only analysis (DTOs + hint engine); zero omac deps, insulated from schema churn.
- Volatile concerns injected / single-sourced: `audit` owns read+write of its format (`audit.ReadLog`); `netproxy.MatchDomainList` exported and injected (no re-implemented matching); path resolution reuses `audit.EffectivePath`.
- `internal/cli/diagnose.go` (analysis + render) and `diagnose_probe.go` (static probe) are thin composition roots that adapt concrete types into the DTOs.

## Verification

- `go test ./...` green; `gofmt`/`go vet` clean.
- Behavioral tests: audit reader (round-trip + malformed-line tolerance), every hint rule, focused-vs-`-v` rendering, and all static probe outcomes (proxy allow/deny/prompt, raw-TCP, loopback, hard-deny).
- Exercised end-to-end against real sandboxed failures: blocked npm/pypi repo, allow∧deny clash, gradle proxy-unaware, kubectl `:6443`, docker socket, git-over-SSH, loopback services.

## Notes for review

- Suggested reading order: `internal/diagnose/diagnose.go` (the model) → `internal/audit/read.go` → `internal/cli/diagnose.go` → `diagnose_probe.go`.
- **Scope note:** an earlier revision also shipped an opt-in `--probe --live` sandboxed-connection check (spawned a real sandbox + temp-file IPC + watchdog) and a `doctor` bridge-port-in-use check. Both were dropped to keep the PR to a single, clearly-defined value: read-only diagnosis with no side effects. The static `--probe` already gives an authoritative verdict by reusing the real filter, so `--live` added machinery for marginal gain.


internal/cli/diagnose.go
```diff
@@ -0,0 +1,355 @@
...
+	fs.BoolVar(verbose, "v", false, "Alias for --verbose.")
+	if err := fs.Parse(reorderFlagsFirst(args)); err != nil {
+		return ExitMisuse
+	}
+	if *runSel != "last" && *runSel != "all" {
+		fmt.Fprintln(env.Stderr, "omac diagnose: --run must be last|all")
+		return ExitMisuse
+	}
+
+	profile, profPath, err := sandboxprofile.Resolve(*profileRef)
```
NoRiceToday: [BLOCKER] bug

`runDiagnose` calls the mutating `sandboxprofile.Resolve`, which scaffolds `~/.config/omac/sandbox-profiles/default.json` on a fresh install. The PR describes diagnose as "pure, read-only inspection — never mutates state," but this writes a profile file to disk.

Repro: `HOME=$(mktemp -d) omac diagnose` → `default.json` appears.

The sibling advisory `doctorSandboxProfileWarnings` correctly uses `ResolveReadOnly`; diagnose should too.

```suggestion
	profile, profPath, err := sandboxprofile.ResolveReadOnly(*profileRef)
```

internal/cli/doctor.go
```diff
@@ -147,12 +149,39 @@ func runDoctor(args []string, env *Env) int {
...
 	}
 	return ExitOK
 }
 
+// doctorProfileLint runs the profileaudit static linter over the resolved
+// sandbox profile and prints its findings as advisory warnings. It closes
+// the gap where doctor never surfaced the security lint (previously only
+// reachable via `omac provenance --check`).
+func doctorProfileLint(env *Env, profileRef string) {
+	profile, _, err := sandboxprofile.Resolve(profileRef)
```
NoRiceToday: [BLOCKER] bug

`doctorProfileLint` calls `sandboxprofile.Resolve(profileRef)`, which scaffolds `default.json` when missing. This breaks the previously-passing `TestDoctorExistingProfileUnchanged`, which asserts doctor does not mutate a fresh home. Verified: passes on `origin/main`, fails on this branch.

The function is advisory and reads the profile only to lint it — it must not write.

```suggestion
	profile, _, err := sandboxprofile.ResolveReadOnly(profileRef)
```

internal/cli/diagnose.go
```diff
@@ -0,0 +1,355 @@
...
+	view := diagnoseView{
+		Version:     env.Version,
+		OS:          osinfo.Detect().String(),
+		Workdir:     env.Workdir,
+		ProfileName: profile.Meta.Name,
+		ProfilePath: profPath,
+		Policy:      pol,
+		RunScope:    *runSel,
+		RunID:       runID,
+		ExitCode:    runExitCode(events, runID),
```
NoRiceToday: [SHOULD FIX] bug / api-contract

Under `--run all`, `runID=""`, so `runExitCode` returns the last `session.stop` exit code regardless of run. The text renderer suppresses it (`v.RunScope != "all"` at diagnose.go:218), but the JSON renderer emits `exit_code` whenever non-nil — so `--json --run all` reports one run's exit code under a scope labeled "all runs." Misleading for a "shareable bundle."

Fix: null out `view.ExitCode` when `*runSel == "all"` before rendering, so JSON and text agree.

internal/cli/diagnose_test.go
```diff
@@ -0,0 +1,200 @@
...
+	drain()
+
+	if code != ExitOK {
+		t.Fatalf("code=%d", code)
+	}
+	var got struct {
+		Report struct {
+			Denied  int `json:"denied"`
+			Blocked []struct {
+				Host string `json:"Host"`
```
NoRiceToday: [SHOULD FIX] maintainability / missing-test

The decode tag is `json:"Host"` (capital H), but `BlockedHost.Host` is emitted as `json:"host"` (lowercase) — see diagnose.go:52. The mismatch means `got.Report.Blocked[0].Host` is always `""`. The test only asserts `Denied == 1`, never the host value, so it passes but proves nothing about the blocked-host JSON contract. A real JSON-shape regression would go undetected.

```suggestion
				Host string `json:"host"`
```

After fixing the tag, add an assertion: `got.Report.Blocked[0].Host == "a.example"`.

NoRiceToday:
This is ready to merge from my side.

PR #138, opened 2026-07-21
feat(start): --auto-register-skills for workdir-local skills with no required values

### LINKED ISSUE #136: feat(start): auto-register workdir-local skills with no required secrets/fields on `omac start`
## What

Add a `--auto-register-skills` flag to `omac start` that silently registers discovered workdir-local skills whose `omac.yaml` declares no required secrets and no required config fields, instead of refusing to start. Skills with required secrets/fields still surface via the existing `findUnregisteredSkills` gate so the user is prompted for values.

## Why

`omac start` hard-fails with `ExitPrerequisiteMissing` whenever a discovered skill directory has an `omac.yaml` but no entry in `.opencode/sidecar.json`. The registry is deliberately gitignored (commit `c5ef7cd` — per-environment state), so every fresh clone / worktree / Superset workspace of a repo that ships committed skill bundles forces an interactive `omac register <skill>` before the agent can start.

This bites the omac repo itself (ships `.opencode/skills/{echo-rest,self-audit}` with `omac.yaml`, registry gitignored) on every fresh workspace. It also bites any downstream repo that ships example/test skills the same way. `omac serve` already has an `autoRegister` path (`serve.go:1206`); `omac start` does not.

The current workaround — running `omac register` manually each time — is friction the user has hit repeatedly and tried to fix before. A flag mirrors serve's behaviour and lets launchers (e.g. the personal `oco` wrapper) opt in without forcing interactive prompts on every workspace.

## How

- Add `--auto-register-skills` to the `omac start` / `omac continue` / `omac resume` shared launch flags (`launchOpts` + `parseLaunchArgs`).
- Before the `findUnregisteredSkills` gate (`start.go:~314`), when the flag is set: scan discovered workdir-local skills (Kind == "workdir"), and for each that is absent from the registry AND whose `omac.yaml` sidecar declares no required secrets and no required config fields, write a registry entry mirroring `serve.go`'s `autoRegister` (bundle hash, declared secret names, relative SkillDir under the workdir).
- Skills WITH required secrets/fields are NOT auto-registered — they still surface via the gate so the user is prompted. Auto-registration never silently skips a registration prompt that would otherwise have asked for a value.
- After auto-registering, reload the workdir registry and re-merge so the gate and the rest of `runLaunch` see the freshly-registered skills.
- Scope is workdir-local only; user-global skills continue to be registered once, globally, via `omac register --global`.

## Non-goals

- Changing the gitignore policy for `.opencode/sidecar.json` (stays per-environment by design).
- Auto-registering skills with required secrets/fields (those still prompt).
- Backfilling the Superset `SUPERSET_*` routing vars into the compiled-in `DefaultAllowVars` — that stays a per-user profile edit since Superset is a single-user deployment for now (tracked separately in the user's personal `tng-customized.json`).

## Verification

- New tests in `internal/cli/start_autoregister_test.go`: eligibility predicate (no sidecar / optional-only / required secret / required field), the auto-register flow (only eligible skills registered, required-value skills still surfaced by the gate), already-registered skip, user-global skip.
- `gofmt -l` clean, `go vet` clean.
- `go test ./internal/cli/ ./internal/sandboxprofile/` — all new + existing tests pass (one pre-existing sandbox-artifact failure, `TestDoctorHarnessBinarySection`, fails identically on the clean tree because it reads the user's real `~/.config/omac/config.yaml`).
- `omac start --help` shows the new flag.

NoRiceToday:
Closes #136

## What

Add a `--auto-register-skills` flag to `omac start` (and the shared `start`/`continue`/`resume` launch pipeline) that silently registers discovered workdir-local skills whose `omac.yaml` declares no required-and-unsatisfiable secrets or config fields, instead of refusing to start with `ExitPrerequisiteMissing`.

## Why

`omac start` hard-fails whenever a discovered skill directory has an `omac.yaml` but no entry in `.opencode/sidecar.json`. The registry is deliberately gitignored per-environment state (commit `c5ef7cd`), so every fresh clone / worktree / Superset workspace of a repo that ships committed skill bundles forces an interactive `omac register <skill>` before the agent can start.

This bites the omac repo itself (ships `.opencode/skills/{echo-rest,self-audit}` with `omac.yaml`, registry gitignored) on every fresh workspace, and any downstream repo that ships example skills the same way. `omac serve` already has an `autoRegister` path (`serve.go`); `omac start` does not. Launchers like the personal `oco` wrapper can opt in without forcing prompts on every workspace.

## How

- Add `--auto-register-skills` to the shared launch flags (`launchOpts` + `parseLaunchArgs`), inherited by `start`/`continue`/`resume`.
- Before the `findUnregisteredSkills` gate, when the flag is set: scan discovered workdir-local skills (Kind == \"workdir\"), and for each absent from the registry AND whose `omac.yaml` sidecar is fully satisfiable at start time without prompting, write a registry entry mirroring `serve.go`'s `autoRegister` (bundle hash, declared secret names, relative SkillDir).
- \"Satisfiable without prompting\" mirrors `runLaunch`'s resolution precedence (`start.go`):
  - a config field is satisfiable if it has a `default:`, a `default_from_env:`, or is explicitly `required: false` (a field with a default but no explicit `required: false` is still satisfiable — `IsRequired()` returns true for nil `Required`, but the default fills the value without user input);
  - a secret is satisfiable if it is in `env_passthrough` (host env supplies it at runtime) or is explicitly `required: false`.
- Skills with at least one required-and-unsatisfiable secret/field are NOT auto-registered — they still surface via the gate so the user is prompted. Auto-registration never silently skips a registration prompt that would otherwise have asked for a value.
- After auto-registering, reload the workdir registry and re-merge so the gate and the rest of `runLaunch` see the freshly-registered skills.
- Scope is workdir-local only; user-global skills continue to be registered once, globally, via `omac register --global`.

## Verification

- `gofmt -l` clean, `go vet` clean.
- `go test ./internal/...` — all packages green.
- New tests in `internal/cli/start_autoregister_test.go` (10 tests): eligibility predicate (no sidecar / optional-only / required secret / required field no default / required field with default → eligible / required secret in env_passthrough → eligible), the auto-register flow (only eligible skills registered, required-value skills still surfaced by the gate, has-default skill registered end-to-end), already-registered skip, user-global skip.
- `omac start --help` shows the new flag.
- Manual end-to-end: `oco --prompt 'hi'` in a fresh workdir of the omac repo auto-registers both `echo-rest` and `self-audit` and launches without the refusal.

## Non-goals

- Changing the gitignore policy for `.opencode/sidecar.json` (stays per-environment by design).
- Auto-registering skills with required-and-unsatisfiable secrets/fields (those still prompt).
- Backfilling the Superset `SUPERSET_*` routing vars into the compiled-in `DefaultAllowVars` — that stays a per-user profile edit since Superset is a single-user deployment for now (tracked in the user's personal `tng-customized.json`, not in this repo).

🤖 Generated with opencode

### DIFF
--- a/CREATING_A_SKILL.md
+++ b/CREATING_A_SKILL.md
@@ -596,9 +596,13 @@ checked:
 2. **Unregistered skill on disk** (a directory with an `omac.yaml` under
    *either* skill source — the workdir-local layer or the user-global
    layer, see §2 — that omac has never seen). Refuses to start;
-   prints the exact register command for each one. Re-registration is
-   mandatory because it's the only point at which secret-prompting and
-   config-prompting happen.
+   prints the exact register command for each one. Standard `omac start`
+   always takes this path. The explicit `--auto-register-skills` start-family
+   opt-in silently registers only workdir-local skills whose required config
+   and secrets resolve without prompting; it never bypasses an unresolved
+   required value. All other unregistered skills remain rejected with their
+   registration command. Explicit registration remains the normal path for
+   config and secret prompting.
 
 3. **Bundle hash drift** (the registered skill's source files changed
    since register). The bundle hash covers `omac.yaml` and every
--- a/README.md
+++ b/README.md
@@ -888,7 +888,12 @@ omac [--workdir <dir>] <subcommand> [flags] [args]
                roots (workdir-local .agents/skills + .opencode/skills,
                plus the user-global layers), or if a registered skill's
                bundle changed since register, or if a required config
-               field is unresolvable. Auto-deregisters
+               field is unresolvable. `--auto-register-skills` is an
+               opt-in for this start-family launch: it silently registers
+               only workdir-local skills whose required config and secrets
+               resolve without prompting. Other unregistered skills still
+               refuse launch and print their registration command.
+               Auto-deregisters
                (silently) skills whose directory has vanished; secrets +
                config persist for safety. Flags:
                  --sandbox <profile>     pick a sandbox profile
@@ -901,6 +906,8 @@ omac [--workdir <dir>] <subcommand> [flags] [args]
                                          scope; removed on exit
                  --keep-running          don't stop sidecars on exit
                  --accept-skill-changes  tolerate bundle_hash drift
+                 --auto-register-skills  silently register eligible
+                                         workdir-local skills only
                  --skip-secret-pattern   don't enforce a secret's pattern
                                          on an env_passthrough value
                  --verbose               lifecycle logging (prints cache
--- a/internal/builtinskills/assets/omac-write-a-skill/references/creating-a-skill.md
+++ b/internal/builtinskills/assets/omac-write-a-skill/references/creating-a-skill.md
@@ -596,9 +596,13 @@ checked:
 2. **Unregistered skill on disk** (a directory with an `omac.yaml` under
    *either* skill source — the workdir-local layer or the user-global
    layer, see §2 — that omac has never seen). Refuses to start;
-   prints the exact register command for each one. Re-registration is
-   mandatory because it's the only point at which secret-prompting and
-   config-prompting happen.
+   prints the exact register command for each one. Standard `omac start`
+   always takes this path. The explicit `--auto-register-skills` start-family
+   opt-in silently registers only workdir-local skills whose required config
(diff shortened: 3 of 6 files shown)


nhuelstng:
lgtm

nhuelstng (approved):

PR #144, opened 2026-07-21
fix(toolcache): persist harness plugin cache across workdirs

### LINKED ISSUE #143: Sandbox cache isolation breaks harness plugin persistence across workdirs (cold cache / offline)
### What

Since the cache-isolation change (#104), a harness's own plugins/extensions are re-fetched in every new workdir instead of being installed once. In the happy path this is a redundant per-project re-download; when the re-fetch can't complete inside the sandbox, the plugin fails to load. omac and the harness still launch — it is specifically the harness's plugin cache that regressed.

### Why

XDG-aware harnesses resolve their cache root to `$XDG_CACHE_HOME/<harness>` and install config-declared plugins under `$XDG_CACHE_HOME/<harness>/packages` (verified against OpenCode `global.ts` → `cache = xdgCache/opencode`, `npm.ts` → `directory = <cache>/packages/<pkg>`). Because #104 scopes `XDG_CACHE_HOME` **per-workdir** (`~/.cache/omac/<sha256("v1:<domain>:<canonical-workdir>")>/xdg`), every new project directory starts with a cold harness cache.

**Baseline effect (always present): a redundant re-fetch.** The harness re-downloads its own user-configured plugins from their registry on the first launch in each new workdir, and the copy installed under the pre-upgrade `~/.cache/<harness>` is orphaned (a grant #104 also removed). This is not about running offline — the harness needs network for the model regardless. The point is that user-controlled state, which should install once, is re-fetched per project.

**Conditional effect: outright failure**, whenever the re-fetch can't complete inside the sandbox:
- a non-interactive / `serve` launch or a prompt timeout — the compiled-in default profile is `network.mode: filtered` with `network_prompt.on_unavailable: deny`, so the fetch is denied;
- the plugin registry needs a VPN and isn't reachable;
- or the plugin was a locally-placed package rather than a registry dependency, so it is never re-fetched.

The original report (a plugin that stopped working after the upgrade) is one of these conditional cases; the always-present cost is the per-workdir re-fetch.

The harness's plugin cache is populated from the user's harness config, not from untrusted project code, so per-workdir isolation gives no security benefit (the #26 threat model is about build/tool caches). It is also inconsistent with the harness's other state: `~/.local/share/<harness>` and `~/.config/<harness>` are granted globally and not remapped — only the cache had been isolated per-workdir.

### How (proposed)

Split the persistent cache into two scopes:

- **Per-workdir build scope** (`OMAC_CACHE_DIR`) — `GOCACHE`, `GOMODCACHE`, `NPM_CONFIG_CACHE`, `PIP_CACHE_DIR`, `CARGO_HOME`. Stays isolated per project (preserves #26).
- **Per-harness XDG scope** (`OMAC_XDG_CACHE_DIR`, keyed by harness name) — `XDG_CACHE_HOME`. Shared across workdirs so the harness's plugins install once instead of being re-fetched per project.

No global `~/.cache` grant is restored; the harness scope is still an isolated `~/.cache/omac/<hash>` directory, honoring the design rule "must use a documented redirect into the selected isolated cache; must not restore a global harness cache grant." In `--ephemeral-cache` mode both scopes remain the same per-launch directory.

Note: this does not migrate the orphaned pre-upgrade `~/.cache/<harness>` contents — after the fix the plugin is re-fetched once into the new stable scope, then reused everywhere.

### Verification

- Unit: harness scope is workdir-independent and distinct from workdir/serve domains; `EnvironmentSplit` routes XDG vs build caches to their own scopes.
- e2e: a marker written under `$XDG_CACHE_HOME` in project A is visible from project B (shared harness scope), while `OMAC_CACHE_DIR` differs (build cache still isolated).
- Before/after with real pre-fix vs fixed binaries at OpenCode's actual plugin path (`$XDG_CACHE_HOME/opencode/packages`): before, `XDG_CACHE_HOME` differs per workdir and the plugin is lost in project B; after, it is shared and the plugin persists.
- Existing cache-isolation e2e suite (worktree isolation, serve scoping, ephemeral cleanup, env-allowlist override) still passes.

nhuelstng:
## What

Fixes #143. Splits the persistent tool cache into two scopes so a harness's own plugins/extensions install once and persist across workdirs, while project build caches stay isolated per workdir.

- **Per-workdir build scope** (`OMAC_CACHE_DIR`): `GOCACHE`, `GOMODCACHE`, `NPM_CONFIG_CACHE`, `PIP_CACHE_DIR`, `CARGO_HOME` — unchanged, still isolated per project.
- **Per-harness XDG scope** (`OMAC_XDG_CACHE_DIR`, `toolcache.DomainHarness`): `XDG_CACHE_HOME` — keyed by harness name, shared across all workdirs.

## Why

XDG-aware harnesses resolve their cache root to `$XDG_CACHE_HOME/<harness>` and install config-declared plugins under `$XDG_CACHE_HOME/<harness>/packages`. Because #104 scopes `XDG_CACHE_HOME` per-workdir, every new project directory starts with a cold harness cache.

The baseline effect (always present) is a **redundant re-fetch**: the harness re-downloads its own user-configured plugins from their registry on the first launch in each new workdir, and the copy under the pre-upgrade `~/.cache/<harness>` is orphaned (a grant #104 also removed). This is not about running offline — the harness needs network for the model regardless; the point is that user-controlled state that should install once is re-fetched per project.

That re-fetch becomes an **outright failure** only when it can't complete inside the sandbox: a non-interactive / `serve` launch or prompt timeout (default `network.mode: filtered`, `network_prompt.on_unavailable: deny`), a VPN-gated registry, or a locally-placed (non-registry) plugin that is never re-fetched. The original "plugin stopped working after the upgrade" report is one of those conditional cases.

The harness's plugin cache is user-config state, not untrusted project output, so per-workdir isolation gives no security benefit (the #26 threat model is build/tool caches). It also mirrors the harness's other dirs — `~/.local/share/<harness>` and `~/.config/<harness>` are already global and not remapped.

## Security posture — what of #104's hardening is kept vs relaxed

**Kept:**
- No broad host-cache grants: `~/.cache`, `~/Library/Caches`, `~/go` stay ungranted; the removed harness-global grants (`~/.cache/opencode`, …) are **not** restored. The harness scope is still an isolated `~/.cache/omac/<hash>` dir.
- Per-workdir isolation of build/tool caches (`GOCACHE`/`GOMODCACHE`/`NPM_CONFIG_CACHE`/`PIP_CACHE_DIR`/`CARGO_HOME`) — the security-critical part where untrusted project code writes. They have explicit overrides so tools do not fall back to `XDG_CACHE_HOME`.
- Injected env still overrides host-inherited values (hostile `GOCACHE`/`CARGO_HOME` replaced), and the sandbox re-exec grant-checks **both** cache dirs (`grantedCacheDir`) so neither bypasses the filesystem allowlist.
- `--ephemeral-cache` collapses both scopes to a per-launch temp dir (full isolation); `~/.rustup`/`~/.cargo/bin` stay read-only.

**Relaxed (deliberate):**
- `XDG_CACHE_HOME` is now shared across a user's workdirs for a given harness. Intended writer: the harness's own plugin cache (user-controlled). Residual surface: any *other* XDG-aware tool a project invokes that writes under `$XDG_CACHE_HOME` and lacks a dedicated redirect would share that entry across the same user's same-harness projects — a narrow cross-project surface per-workdir XDG previously closed.
- Not narrowed to `$XDG_CACHE_HOME/<harness>` only, because OpenCode derives its whole cache root from `XDG_CACHE_HOME` with no env to relocate just its subtree (all-or-nothing on that variable).

## How

- `internal/toolcache`: add `DomainHarness`, `DescribeHarness`/`PrepareHarness`, split `Environment` into `EnvironmentSplit(buildDir, xdgDir, mode)` (adds `OMAC_XDG_CACHE_DIR`).
- `internal/cli/start.go`, `serve.go`: prepare both scopes, grant both leaves read+write, inject `OMAC_XDG_CACHE_DIR` alongside `OMAC_CACHE_DIR`.
- `internal/sandboxrun/run.go`: re-derive and grant-check both dirs independently on the sandbox re-exec (factored into `grantedCacheDir`).
- `internal/sandboxbrief`, `provenance.go`: reflect the split in the agent briefing and diagnostics.
- `--ephemeral-cache` and `--no-sandbox` behaviour unchanged: both scopes collapse to the single per-launch dir (or none).
- Does not migrate the orphaned pre-upgrade `~/.cache/<harness>` contents: the plugin is re-fetched once into the new stable scope, then reused.
- Docs: `NONO_SANDBOX.md` and the cache-isolation design doc updated.

## Verification

- **Unit** (`internal/toolcache`, `internal/cli`, `internal/sandboxbrief`): harness scope is workdir-independent and distinct from workdir/serve domains; `EnvironmentSplit` routes XDG vs build caches to their own scopes; serve holds/clears both scope locks.
- **e2e** (`internal/e2e`, bubblewrap host): new `TestE2ECacheHarnessXDGSharedAcrossWorkdirs` — a marker under `$XDG_CACHE_HOME` in project A is readable from project B (shared harness scope) while `OMAC_CACHE_DIR` differs (build cache still isolated). The `-run '^TestE2ECache'` suite passes except `TestE2ECacheTools`, which fails identically on the base commit due to a host Python-3.14 `sysconfig` path (unrelated; passes in CI). `-tags=e2e_fast` passes.
- **Before/after** with real pre-fix vs fixed binaries at OpenCode's actual plugin path: before, `XDG_CACHE_HOME` differs per workdir and the plugin is lost in project B; after, it is shared and the plugin persists.

### DIFF
--- a/docs/NONO_SANDBOX.md
+++ b/docs/NONO_SANDBOX.md
@@ -74,6 +74,9 @@ environment:
     - CARGO_HOME
 ```
 
+If an external nono profile sets `environment.allow_vars`, it must also
+include `OMAC_XDG_CACHE_DIR` (covered by the `OMAC_*` glob above).
+
 External nono profiles do not receive the built-in sandbox re-exec's trusted
 cache reinjection. Omitting a redirect lets that tool use its default cache
 location; omac does not provide an automatic substitute.
@@ -97,14 +100,24 @@ launcher-config schema. Available placeholders: `{{socket}}`,
 
 ### Tool-cache redirection and external nono-profile limitations
 
-For sandboxed launches, omac prepares a per-scope tool-cache directory
-(`~/.cache/omac/<sha256(identity)>` for persistent mode, a per-launch
-temp dir for `--ephemeral-cache`) and redirects the caches supported by
-`XDG_CACHE_HOME`, `GOCACHE`, `GOMODCACHE`, `NPM_CONFIG_CACHE`,
-`PIP_CACHE_DIR`, and `CARGO_HOME`. It injects `OMAC_CACHE_DIR` /
-`OMAC_CACHE_MODE` plus those six redirects into the sandbox runtime's
-process environment, then grants only that scope leaf read+write in
-the sandbox argv. Unsupported third-party tools that hardcode another
+For sandboxed launches, omac prepares tool-cache directories under
+`~/.cache/omac/<sha256(identity)>` (persistent mode) or a per-launch temp dir
+(`--ephemeral-cache`) and redirects the caches supported by `XDG_CACHE_HOME`,
+`GOCACHE`, `GOMODCACHE`, `NPM_CONFIG_CACHE`, `PIP_CACHE_DIR`, and `CARGO_HOME`.
+
+Persistent mode uses **two** scopes. The build caches (`GOCACHE`, `GOMODCACHE`,
+`NPM_CONFIG_CACHE`, `PIP_CACHE_DIR`, `CARGO_HOME`) live in a **per-workdir**
+scope (`OMAC_CACHE_DIR`) — untrusted project code writes here, so it stays
+isolated per project. `XDG_CACHE_HOME` lives in a **per-harness** scope
+(`OMAC_XDG_CACHE_DIR`), keyed by harness name and shared across every workdir:
+the harness's own cache holds the plugins/extensions declared in the user's
+harness config, which are user-controlled and installed once rather than
+re-fetched from a registry in every project directory (a per-project re-fetch
+the sandbox network policy may block). In `--ephemeral-cache` mode both scopes
+are the same per-launch directory. omac injects `OMAC_CACHE_DIR`,
+`OMAC_XDG_CACHE_DIR`, and `OMAC_CACHE_MODE` plus the six redirects into the
+sandbox runtime's process environment, then grants only those scope leaves
+read+write in the sandbox argv. Unsupported third-party tools that hardcode another
 cache path need explicit profile configuration; omac does not redirect
 them automatically. `~/.rustup` may be read-only runtime-visible for
 toolchain execution, but it is not a writable or cache grant. See the
--- a/docs/superpowers/specs/2026-07-10-cache-isolation-design.md
+++ b/docs/superpowers/specs/2026-07-10-cache-isolation-design.md
@@ -131,13 +131,14 @@ The inner harness receives these injected values, overriding inherited values:
 
 | Variable | Isolated location |
 |---|---|
-| `XDG_CACHE_HOME` | `<cache>/xdg` |
+| `XDG_CACHE_HOME` | `<xdg-cache>/xdg` |
 | `GOCACHE` | `<cache>/go-build` |
 | `GOMODCACHE` | `<cache>/go-mod` |
 | `NPM_CONFIG_CACHE` | `<cache>/npm` |
 | `PIP_CACHE_DIR` | `<cache>/pip` |
 | `CARGO_HOME` | `<cache>/cargo` |
 | `OMAC_CACHE_DIR` | `<cache>` |
+| `OMAC_XDG_CACHE_DIR` | `<xdg-cache>` |
 | `OMAC_CACHE_MODE` | `persistent` or `ephemeral` |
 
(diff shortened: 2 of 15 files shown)


nhuelstng:
Brainstorm wether a per-config cache is not the better solution

PR #148, opened 2026-07-21
feat(netproxy): explain blocked package-registry fetches

nhuelstng:
## What

When the sandbox network filter blocks — or fails to reach — a package-registry host, append an actionable, reason-tailored note so the user (and the agent) understands *why* a plugin/provider fetch failed, instead of seeing a generic denial and an unexplained missing plugin.

Related to #143 (a harness fetching a config-declared plugin into a cold cache), but independent and useful on its own — it helps diagnose any blocked registry fetch.

## Why

A harness installing a config-declared plugin/provider fetches from its registry (npm/pypi/crates/…). Two distinct failure paths currently give the user no legible cause:

- **Policy denial** (HTTP 403, `denyBody`): under the default profile (`network.mode: filtered`, `network_prompt.on_unavailable: deny`) the fetch is denied when there is **no interactive prompt to show** (`serve`/headless), when `network_prompt` is disabled, or on prompt timeout. The user just sees "the plugin stopped working".
- **Connection failure** (HTTP 502, `upstreamErrorBody`): the host was allowed but unreachable — most often a private/VPN-scoped registry with the VPN down. This never went through the deny path, so a deny-oriented message wouldn't help.

## How

- `registry.go`:
  - `isPackageRegistry(host)` — curated public registries (npm/yarn/pypi/pythonhosted/crates/rubygems/goproxy/GitHub Packages) plus a heuristic for private registries (a `registry`/`npm` host label, or an ecosystem name), so a VPN-scoped `registry.<corp>` is still recognized. Used only to enrich an already-failing request, so a false positive is harmless.
  - `registryDenyHint(host, reason)` — returns "" for non-registries; otherwise a note **tailored to the deny reason**, because the remedy differs: no interactive prompt available (`serve`/headless) → pre-allow via `network.allow_domain`; prompt disabled → add to `allow_domain`; declined at prompt → re-run and Allow; deny rule → remove it.
  - `registryUpstreamHint(host)` — connection-failure note pointing at VPN/reachability.
- `server.go`: `denyBody` appends the deny hint (single hook — covers both the CONNECT and plain-HTTP forward paths); the upstream-error branch logs the connection hint.

### Consequence for the user

Before: opencode logged a generic plugin-install failure and the sandbox logged a generic denial — no link between them. After: the denial body / diagnostics name the registry, state the specific reason (and whether a prompt could even have been shown), and give the exact knob to fix it. The failure becomes self-explanatory, and the VPN case is called out separately from the policy case.

## Verification

`internal/netproxy` unit tests: registry matcher (public hosts, private-registry shapes, negatives like `api.anthropic.com`/`github.com`); deny/upstream hints fire only for registries; the deny remedy text is tailored per reason; and `denyBody` appends the hint for a registry host but not otherwise. `go build ./...`, `go vet`, `go test ./internal/netproxy/... ./internal/sandboxrun/...` all pass.

### DIFF
--- a/internal/netproxy/registry.go
+++ b/internal/netproxy/registry.go
@@ -1,0 +1,95 @@
+package netproxy
+
+import (
+	"fmt"
+	"strings"
+)
+
+// knownRegistrySuffixes are well-known public package-registry hosts. They are
+// matched as exact hosts or dot-suffixes (so "foo.pypi.org" also matches).
+var knownRegistrySuffixes = []string{
+	"registry.npmjs.org",
+	"registry.yarnpkg.com",
+	"npm.pkg.github.com",
+	"pypi.org",
+	"files.pythonhosted.org",
+	"crates.io",
+	"static.crates.io",
+	"rubygems.org",
+	"proxy.golang.org",
+	"sum.golang.org",
+}
+
+// isPackageRegistry reports whether host looks like a package registry. It is a
+// heuristic used only to enrich an already-emitted denial/error message, so a
+// false positive is harmless (it appends a "looks like a registry" note to a
+// request that was already going to fail). Beyond the curated public hosts it
+// also recognizes common private-registry shapes (a "registry"/"npm" label, or
+// an ecosystem name in the host), so a VPN-scoped `registry.<corp>` still hits.
+func isPackageRegistry(host string) bool {
+	host = strings.ToLower(strings.TrimSuffix(host, "."))
+	for _, s := range knownRegistrySuffixes {
+		if host == s || strings.HasSuffix(host, "."+s) {
+			return true
+		}
+	}
+	labels := strings.Split(host, ".")
+	for _, l := range labels {
+		switch l {
+		case "registry", "npm":
+			return true
+		}
+	}
+	for _, frag := range []string{"pypi", "pythonhosted", "crates", "rubygems", "goproxy"} {
+		if strings.Contains(host, frag) {
+			return true
+		}
+	}
+	return false
+}
+
+// registryDenyHint returns an actionable note appended to a policy-denial body
+// when the denied host looks like a package registry — the common shape of a
+// harness fetching a config-declared plugin/provider into a cold cache. The
+// guidance is tailored to why the request was denied (reason), because the
+// remedy differs: no interactive prompt was available vs. the prompt is
+// disabled vs. the user declined vs. an explicit deny rule. Returns "" for
+// non-registry hosts so callers can append unconditionally.
+func registryDenyHint(host, reason string) string {
+	if !isPackageRegistry(host) {
+		return ""
+	}
+	var next string
+	switch {
+	case strings.Contains(reason, "prompt unavailable"):
+		next = "This launch has no interactive network prompt (e.g. `omac serve` or a non-TTY\n" +
+			"session), so the request was auto-denied (on_unavailable=deny). Pre-allow it with\n" +
+			"network.allow_domain, or pre-warm the cache once via an interactive `omac start`."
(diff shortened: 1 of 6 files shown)


nhuelstng:
## Review follow-up — applied

Reviewed by driving the real `netproxy.Server` over CONNECT and plain-HTTP and asserting on the bytes a client receives, rather than on the pure functions. That surfaced two defects the hint functions' own tests could not see, plus three smaller ones. All fixed test-first; three commits pushed.

### What

**1. The deny hint's wrapping was broken in every reason branch.** The pre-wrapped remedy was interpolated mid-sentence, producing a 134-char line in a body whose prose is otherwise ~80:

```
install likely tried to fetch from it into a cold cache. This launch has no interactive network prompt (e.g. `omac serve` or a non-TTY
```

**2. The connection-failure hint never reached the client.** It went only to `s.logf`, and `newDiagSink` writes to `~/.local/state/omac/sandbox.log` whenever stderr is a terminal — so the VPN-down case the hint exists to explain was invisible to the agent that made the request. A real CONNECT to an allowed-but-unreachable registry returned only:

```
HTTP/1.1 502 Bad Gateway

upstream dial failed: dial tcp 10.1.2.3:443: connect: connection refused
```

**3. `prompt:needs_intent` carried no hint** — `denyBody` returned early, and the reason would have fallen through the hint's switch to "remove a deny rule", the wrong remedy for a path whose fix is the intent round-trip.

**4. The DNS-failure body contradicted itself** — it claimed `DENIED BY THE SANDBOX network policy` and listed three policy knobs, while the hint below correctly said nothing had been denied by policy.

**5. Hard-deny bodies offered `allow_domain`, which cannot override them.** Verified against `Filter.Check` with both `metadata.google.internal` and `169.254.169.254` in `allow_domain` — both still return `hard-deny metadata host`, because the guard runs before any rule. This one was outside the original review scope but is the same defect class as 4.

### Why

The feature's value is that a blocked fetch explains itself. A hint that wraps badly, or that lands in a file the requesting agent cannot read, or that names a knob which provably changes nothing, does not deliver that.

### How

- `registryhint.go` — remedy on its own line; `needs_intent` case added; `flattenForLog` for the log form; DNS and hard-deny hints trimmed to what the new bodies don't already say.
- `server.go` — `upstreamFailureResponse(host, err)` puts the hint in the 502 body and de-duplicates a branch that was repeated verbatim across `handleConnect` and `handleForward`; `denyBody` now switches on reason class (intent / resolution failure / built-in guard / policy).
- `registry.go` → `registryhint.go`: "registry" in this repo already means `internal/registry`, the skill registry.
- `openspec/changes/native-sandbox/specs/sandbox-network/spec.md` — the requirement said every filtered denial must name the policy knobs, which is wrong for the two verdicts that are not policy decisions. Now scoped by class, with four scenarios, plus a note that guidance the child needs belongs in the response body.

Real policy denials are byte-identical, pinned by a counterweight test over all five policy reasons.

### Verification

`go build ./...`, `go vet ./...`, `gofmt -l`, `go test ./...` clean; `-race` clean on `netproxy` and `sandboxrun`. 75 tests in `netproxy`. Pre-existing `TestServer502OnUpstreamFailure` / `TestServer502OnUpstreamDialFailure` pass unchanged. The e2e net-denial assertions (`security_assertions.go:104`, `provenance_test.go:218`) are unaffected — OR-lists that still match on `403`, and they already treat DNS failure as a distinct class.

New tests assert on client-received bytes: hint present in the 502 body, no body line over 100 chars, deny hint within that width across all eight reason branches, logged form single-line (the diag sink appends whole lines and prepends per-line columns), and `allow_domain` cannot rescue a hard-denied host.

PR #157, opened 2026-07-21
feat(intent): deliver explain-more hint inline on deny

nhuelstng:
## What

When the sandbox denies a request, the remediation hint is now delivered **in the deny response itself** — inline in the body *and* on an `X-Omac-Intent-Hint` response header — so the agent no longer needs a follow-up `GET /sandbox/intent` to learn what to do next. The deny channel is **state-aware** and mirrors the GET lookup's state machine.

## Why

- Previously the deny body carried only a short ad-hoc note; the authoritative hint lived solely behind the `GET` lookup, forcing a second round-trip.
- HTTPS/CONNECT clients **discard the deny body** entirely, so the hint now also rides on a response header — the only in-band channel such tooling can surface.
- The hint must reflect the actual denial state, not just `needs_intent`: a plain policy deny that already has a declared intent on file means the user reviewed the reason and declined, so the agent should be told **not** to retry.

## How

- **`internal/intent/hints.go`** (new): the three hint strings are exported constants (`HintUndeclared` / `HintDeclared` / `HintExplainMore`) — one authoritative wording shared by the facade GET lookup and the netproxy deny path. Values are single-line ASCII by contract (valid HTTP header field-values).
- **`HintExplainMore` reworded** (behavioral): now demands a *specific* reason (why / what / how) and states explicitly *"do not retry without an intent explanation."* Because the facade aliases the same constant, this also changes the `GET /sandbox/intent` explain-more response body.
- **`internal/netproxy/server.go`** — the deny channel is now state-aware via `intentHintFor(v)`, mirroring the GET state machine (this replaces the earlier "scoped strictly to `needs_intent`" behavior):
  - `needs_intent` (user clicked "Explain more") → `HintExplainMore`;
  - `prompt:deny` **with** a declared reason on file → `HintDeclared` ("user reviewed and declined — do not retry"); the body also echoes the agent's declared reason;
  - anything else (undeclared / policy / allowlist deny) → no hint.
  `denyHeaders` adds `X-Omac-Intent-Hint` for the first two cases; `denyBody` embeds the same state-aware hint inline. Only the static hint constants reach the header — the agent-supplied reason is body-only and CR/LF-sanitized.
- **`internal/facade/facade.go`**: local consts alias `intent.*`. **POST `/sandbox/intent` now clears the one-shot explain-more flag** (`ClearExplainMore`) after recording, so inline recovery (read hint → POST fuller intent → retry) retires the flag just as a GET-consume would. Without this, a stale flag would revive the "re-declare and retry" hint on a later GET fallback after the user had already declined the retry.
- **`internal/intent/registry.go`**: new `ClearExplainMore(target)` — normalized, idempotent, nil-safe.

## Verification

- `go build ./...` — OK
- `go test ./internal/netproxy ./internal/facade ./internal/intent` — pass
- `go vet` + `gofmt -l` — clean
- Tests: `TestDenyHeaders`, `TestServerNeedsIntentDenyCarriesHintHeader` (wire-level `needs_intent` CONNECT denial carries the header on the parsed 403), `TestIntentHintHeaderIsValidFieldValue` (ASCII/CRLF guard over all three constants), strengthened `TestDenyBodyNeedsIntent`; and for the flag lifecycle `TestIntentPostClearsExplainMore` + `TestRegistryClearExplainMore`.

## Header-injection note

The only value inserted into the header is a compile-time hint constant; `verdict.Reason` is likewise an internal constant. No external/attacker-controlled data reaches the header, and the ASCII/single-line contract is enforced by test. The agent-supplied intent reason appears only in the deny body, CR/LF-sanitized.


internal/netproxy/server.go
```diff
@@ -30,19 +32,60 @@ var connectTimeout = 30 * time.Second
...
+//     regardless of whether a reason was already on file;
+//   - a plain deny with a reason on file -> the user reviewed a declared
+//     intent and declined it, so tell the agent not to retry;
+//   - anything else (undeclared deny, policy/allowlist deny) -> no hint,
+//     since inviting a retry there would loop or is irrelevant.
+func intentHintFor(v Verdict) string {
+	switch {
+	case strings.Contains(v.Reason, "needs_intent"):
+		return intent.HintExplainMore
+	case strings.Contains(v.Reason, "prompt:deny") && strings.TrimSpace(v.IntentReason) != "":
```
NoRiceToday, line 56: Small documentation-alignment request: the PR description says prompt:deny stays clean and the hint is scoped to needs_intent, but this branch also emits HintDeclared when there is a prior reason. The intent design still describes GET as the sole reliable CONNECT recovery channel. Could we update the PR description and the design record to describe the actual two-channel behavior and this declared-intent case?

NoRiceToday:
Friendly process reminder for future changes: this PR does not have a linked issue. No action is needed for this PR if it was intentionally untracked; please ensure future PRs include their tracking issue or a closing reference such as Fixes #<issue>, so the work remains traceable.

internal/netprompt/prompt.go
```diff
@@ -267,6 +267,9 @@ func (p *Prompter) Prompt(host string, port int) netproxy.PromptResult {
 	}
 	token := labelToToken(label, suffix)
 	result := tokenToResult(token, host, suffix)
+	// The intent was already fetched above to render the dialog; carry it on
+	// the result so a denial can echo it back without another registry read.
+	result.PriorReason = intent
```
NoRiceToday, line 272: Nice to see the in-band path. One lifecycle issue remains: this records the one-shot Explain more marker for the GET fallback, but the new intended recovery flow consumes the inline hint, POSTs a fuller intent, and retries without GET. The marker then remains live. If the user declines that retry and the agent later uses GET as a fallback, the stale marker overrides the declined hint and invites another retry. Could we clear the marker when a replacement intent is posted for this target (or otherwise acknowledge inline recovery), and cover that sequence with a regression test?

internal/netproxy/server.go
nhuelstng: Good catch — the description was stale after `52cc41a` made the deny channel state-aware. Fixed in `30a82f9`:

- **PR description** rewritten to describe the two-channel behavior (inline body + `X-Omac-Intent-Hint` header) and the declared-intent decline case (`prompt:deny` with a reason on file → `HintDeclared`, "do not retry"), replacing the old "scoped strictly to `needs_intent`" wording.
- **Design record** (`docs/superpowers/specs/2026-07-07-intent-declarations-design.md`) updated: GET is now documented as the reliable *fallback*, not the sole CONNECT recovery channel, and the deny channel's state machine (`intentHintFor`) is spelled out.

internal/netprompt/prompt.go
nhuelstng: Confirmed and fixed in `30a82f9`. Traced the lifecycle: the flag was only ever consumed by the GET lookup (`ConsumeExplainMore`), while the inline-recovery flow (read hint → `POST /sandbox/intent` → retry) bypasses GET, so `Record` left the flag live. A later GET fallback after a declined retry would then revive `HintExplainMore` and invite another retry.

Fix:
- New `Registry.ClearExplainMore(target)` (normalized, idempotent, nil-safe).
- `POST /sandbox/intent` calls it after `Record`, so inline recovery retires the flag exactly as a GET-consume would.

Regression tests:
- `TestIntentPostClearsExplainMore` — mark → POST fuller intent → GET now yields the declared hint, not the explain-more re-ask.
- `TestRegistryClearExplainMore` — normalization / idempotency / nil-safety.

NoRiceToday:
Sidenote: given that we decided to keep the spec local, do we want to remove the one committed and changed in this MR? :)

internal/facade/facade.go
```diff
@@ -603,6 +603,12 @@ func (f *Facade) handleSandboxIntent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	f.IntentRegistry.Record(body.Target, body.Reason)
+	// Posting a fuller intent is the in-band answer to an "Explain more" click,
+	// so retire the one-shot flag here. The inline deny hint (body + header)
+	// lets the agent recover without ever hitting the GET lookup that would
+	// otherwise consume it; leaving it live would let a later GET fallback
+	// revive the "re-declare and retry" hint after the user declined the retry.
+	f.IntentRegistry.ClearExplainMore(body.Target)
```
NoRiceToday: One remaining concurrency edge: Record releases the registry mutex before ClearExplainMore acquires it. A concurrent GET /sandbox/intent can run in that gap, consume the old flag, and return HintExplainMore once more even though the replacement intent was posted. Could we make record-and-clear a single atomic registry operation and add a forced-interleaving regression test? The current lifecycle test covers only serialized handler calls.

NoRiceToday:
@nhuelstng from my side, it's ready to be merged now.

Issue #171, opened 2026-07-23
Briefing Injection conflicts with some models


nhuelstng:
Qwen expects system prompt first

GevatterGaul:
Affecting me as well

PR #177, opened 2026-07-27
Feature/codewhale harness

GevatterGaul:
Add the codewhale harness.

There is a workaround for codewhale not having `--append-system-prompt` or similar: We had to add a special mounted codewhale rule.

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -53,7 +53,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        harness: [opencode, claude-code, codex, copilot, pi]
+        harness: [opencode, claude-code, codex, copilot, pi, codewhale]
         # Two omac versions: main (built from source) and release (the latest
         # published binary — what users actually run). The contract stage is
         # identical across both (it's compiled from source), but launch + llm
@@ -62,9 +62,13 @@ jobs:
         omac: [main, release]
         exclude:
           # codex's Rust HTTP client is incompatible with the macOS Seatbelt
-          # sandbox (see e2e.yml / issue #48).
+          # sandbox (see e2e.yml / issue #48). codewhale is excluded on macOS
+          # by analogy (also a Rust CLI with its own HTTP client) — unverified,
+          # re-test and drop if it works.
           - os: macos-latest
             harness: codex
+          - os: macos-latest
+            harness: codewhale
     runs-on: ${{ matrix.os }}
     concurrency:
       group: e2e-smoke-${{ matrix.os }}-${{ matrix.harness }}-${{ matrix.omac }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -38,6 +38,10 @@ on:
         description: 'Pinned pi package spec. Overrides internal/e2e/versions.go for THIS RUN ONLY (the file itself is untouched); ignored if "use latest" is checked'
         type: string
         default: '@earendil-works/pi-coding-agent@0.80.6'
+      codewhale_version:
+        description: 'Pinned codewhale package spec. Overrides internal/e2e/versions.go for THIS RUN ONLY (the file itself is untouched); ignored if "use latest" is checked'
+        type: string
+        default: 'codewhale@0.9.1'
   schedule:
     - cron: '0 8 * * 6' # Saturdays 10:00 CEST
 
@@ -51,15 +55,22 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        harness: [opencode, claude-code, codex, copilot, pi]
+        harness: [opencode, claude-code, codex, copilot, pi, codewhale]
         # codex on macOS is excluded — its Rust HTTP client is incompatible
         # with the macOS Seatbelt sandbox (stream disconnected mid-completion,
         # even with network=open). `omac start codex` on macOS fails loud;
         # --no-sandbox would disable the entire omac sandbox, leaving nothing
         # to assert against. See issue #48.
+        #
+        # codewhale on macOS is excluded by analogy: it too is a Rust CLI with
+        # its own HTTP client (the codex failure class). This is precautionary
+        # and UNVERIFIED for codewhale — re-test on macOS and drop this exclude
+        # (and the matching ones in internal/e2e) if it works.
         exclude:
           - os: macos-latest
             harness: codex
+          - os: macos-latest
+            harness: codewhale
     runs-on: ${{ matrix.os }}
     concurrency:
       group: e2e-${{ matrix.os }}-${{ matrix.harness }}
@@ -82,6 +93,7 @@ jobs:
       E2E_VERSION_CODEX: ${{ github.event.inputs.codex_version }}
       E2E_VERSION_COPILOT: ${{ github.event.inputs.copilot_version }}
       E2E_VERSION_PI: ${{ github.event.inputs.pi_version }}
+      E2E_VERSION_CODEWHALE: ${{ github.event.inputs.codewhale_version }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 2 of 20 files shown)


nhuelstng:
could you add this to the harness tests then too?

GevatterGaul:
> could you add this to the harness tests then too?

Done

nhuelstng:
Reviewed against a local install of `codewhale@0.9.1`; the output below is from that binary and from the PR's own test helpers.

**1. `llm` stage can't authenticate.** CodeWhale rejects ambient `OPENAI_API_KEY` on a custom `base_url`, so `codewhaleConfig`'s provider setup never reaches the model.
```
Error: … Custom endpoint credentials for openai must be bound explicitly.
Ambient provider credentials are not sent to <SKAINET_INTERNAL>.
```
→ add `api_key_env = "OPENAI_API_KEY"` to `[providers.openai]`. Verified: reaches the auth stage, and the token still never touches disk.

**2. `omac resume` / `continue -s <id>` fails.** `--resume` is an `exec`-forwarded flag; at top level `resume` is a subcommand.
```
$ codewhale --resume abc123   → error: unexpected argument   exit=2
$ codewhale resume abc123     → parses (only wants a TTY)
```
→ `ResumeByIDArgs: []string{"resume", id}`, matching codex.

**3. Contract stage fails on the next weekly drift run.** `--model` leads `RunArgs`, so `deriveContract` derives no subcommand and the corpus is top-level `--help` only:
```
tokens=[--model --auto --continue --resume] subcommands=[]
MISSING "--auto", MISSING "--resume"      # both appear only in `exec --help`
```
→ fix (2), plus get `exec` into `Subcommands` (extend `flagsAndSub` past the leading position). The corrected variant resolves every token. `exec` currently has no drift coverage either.

**4. The briefing gets committed into the user's repo.** Replaying the e2e's default prompt (`git add -A && git commit`):
```
in the agent's commit:  .codewhale/rules/omac-sandbox-briefing.md
after cleanup:           D .codewhale/rules/omac-sandbox-briefing.md
```
So the repo keeps the briefing and a staged deletion of a now-tracked file. `defer` also misses SIGKILL. → append the path to `.git/info/exclude` (verified: clean `git status`, nothing committed, user `.gitignore` untouched) plus a stale-file sweep on launch. The workdir write itself is unavoidable: `RULES_DIRS` is workspace-only (`project_context.rs:47`) and global `AGENTS.md`/`instructions.md` are fallbacks skipped whenever the repo defines its own (`:62-78`).

**5. `SkillsBase` in the registry doesn't match what CodeWhale reads.** `codewhale doctor`, in a workdir containing both dirs, never probes `<wd>/.codewhale/skills`:
```
✓ .agents skills dir found at <wd>/.agents/skills
✓ global skills dir found at ~/.codewhale/skills
· selected skills dir: <wd>/.agents/skills
```
`SkillsBase: "codewhale"` makes `WorkdirSkillsDir()` = `.codewhale/skills`, so skills registered under codewhale are invisible to it. → `SkillsBase: SharedSkillsBase`. Verified: `WorkdirSkillsDir` becomes `.agents/skills` while `ConfigHome`/`GlobalSkillsDir` stay `~/.codewhale[/skills]` (they derive from `UserConfigHome`). That also aligns the registry with the e2e's `.agents`, which is correct as written.

**Smaller:** `sandbox_mode = "external-sandbox"` is the semantically right value and is accepted by 0.9.1 · setting approvals/sandbox via the top-level flags instead of `config.toml` would give them drift coverage · the macOS exclude is unverified, one manual `E2E: drift` dispatch would settle it · codewhale is missing from `TestDeriveContractKnownTokens` and `TestHarnessHasServerMode` · `docs/HARNESS_COMPAT.md:3` (and its codex/macOS note), `CREATING_A_SKILL.md:50` and its bundled copy are not updated · no `openspec/changes/support-codewhale-harness/`, unlike claude-code, codex+copilot and pi · DCO is red and no issue is linked · `tableColumns` is a cleaner version of `listCopilotDB`'s inline schema-gate and could replace it.

**Relation to #171 (Qwen expects the system prompt first).** `BriefingFileFunc` is the right direction for that issue and worth linking. #171's cause is that two delivery paths author *message structure* rather than contributing text: opencode's `output.system.push(brief)` (`omac-multidir.ts:292`) adds a second system message, and pi's `event.messages.unshift({role:"system"})` alongside the `systemPrompt` return (`omac-bridge.ts:165-172`) injects it twice. Flag- and file-based delivery let the harness position the text itself, so codewhale is #171-safe by construction. Cheapest first step there: drop pi's redundant `unshift`. It also raises the value of fixing (4) properly — if #171 moves omac toward file-based delivery generally, workdir pollution stops being codewhale-specific.

`go build ./...`, `go test ./internal/...` and `go test -tags=e2e_fast ./internal/e2e/` all pass on the branch.

GevatterGaul:
All five main points and the smaller items are addressed in the follow-up commit. Notably the smoke tier now installs 
  codewhale@0.9.1 and passes the contract stage against the real binary (all 4 tokens resolve), so #2 and #3 are live-verified rather than
  just source-derived.

  1. LLM auth on a custom base_url. Fixed. Added api_key_env = "OPENAI_API_KEY" to [providers.openai], so the token is bound explicitly (and
  still never touches disk — it stays in the process env). Confirmed it now reaches the auth stage. Also switched sandbox_mode to
  external-sandbox, which is the semantically correct value for "running inside an external OS sandbox."

  2. resume is a subcommand, not a flag. Fixed. ResumeByIDArgs → ["resume", id] (matching codex's shape). --continue stays as-is since it is
  a valid top-level flag.

  3. Contract drift. Fixed by making exec lead RunArgs (the model now comes from config.toml's default_text_model), so deriveContract
  captures the subcommand and folds exec --help/resume --help into the corpus — every token resolves. I went this route instead of extending
  flagsAndSub past the leading position: doing that would treat non-leading flag values (--model's value, and the never/external-sandbox
  args if approvals were CLI flags) as subcommands/tokens, which missingTokens would then require in the corpus and fail on. Making exec
  lead gives full coverage without that hazard, and I added codewhale to both TestDeriveContractKnownTokens and TestHarnessHasServerMode.
  Relatedly, I kept approval_policy/sandbox_mode in config.toml rather than top-level flags — production omac start codewhale has no
  RunArgs, so config is the one place that covers both the e2e and interactive paths.

  4. Briefing committed into the repo. Fixed. Before launch the launcher appends the rules-file path to <workdir>/.git/info/exclude
  (idempotent, preserves any existing user entries, never touches the user's .gitignore or the index). Because the entry is written before
  the agent runs, it survives a SIGKILL — the leftover file stays git-ignored until the next run overwrites and cleans it — which closes the
  "staged deletion of a now-tracked file" gap you demonstrated. The on-exit removeBriefingFile stays. New unit tests cover
  append/idempotency/user-entry-preservation and the empty-dir-prune-only cleanup. Agreed the workdir write itself is unavoidable given
  RULES_DIRS is workspace-only.

  5. SkillsBase. Fixed. SkillsBase: SharedSkillsBase, so WorkdirSkillsDir() is .agents/skills (where CodeWhale actually looks) while
  ConfigHome()/GlobalSkillsDir() still resolve to ~/.codewhale[/skills] via UserConfigHome. This also lines the registry up with the e2e's
  .agents.

  Smaller items:
  - sandbox_mode = "external-sandbox" — adopted.
  - tableColumns — extracted as a shared helper and listCopilotDB refactored onto it.
  - TestDeriveContractKnownTokens / TestHarnessHasServerMode — codewhale added to both.
  - Docs — docs/HARNESS_COMPAT.md (harness list + macOS-exclude note) and CREATING_A_SKILL.md (+ its byte-identical bundled copy) updated.
  - openspec/changes/support-codewhale-harness/ — added (proposal + tasks).
  - macOS exclude — still precautionary/by-analogy-to-codex and unverified; I left it in with a comment to drop it after a manual E2E: drift
  dispatch confirms CodeWhale works under Seatbelt.
  - DCO — all four commits are now signed off (rebased); force-pushing.

  #171. Agreed the direction is right — BriefingFileFunc is #171-safe by construction since the harness positions the system text itself.
  Dropping pi's redundant unshift in omac-bridge.ts is a clean separate fix; happy to open it as its own PR/issue rather than fold it in
  here.

PR #179, opened 2026-07-28
feat(netproxy): explain blocked package-registry fetches (pre-release integration)

nhuelstng:
**Issue:** Refs #143, supersedes #148

## What

- Pre-release integration branch: PR #148 plus `main` merged in, validated end to end at one SHA.
- A blocked or failed package-registry fetch now explains itself in the response the client actually receives — deny body **and** the 502 connection-failure body.
- `denyBody` picks its remedy by reason class, so it never offers a knob that cannot change the outcome.
- Drive-by: `scripts/e2e-docker.sh cache` could not run at all (pre-existing, `26899a1`).

## Why

#148 as it stood had two defects that only showed up when the real proxy was driven over CONNECT rather than the hint functions being unit-tested:

- The deny hint spliced its pre-wrapped remedy mid-sentence, producing a 134-char line in a body otherwise wrapped at ~80 — in every reason branch.
- The connection-failure hint went only to `s.logf`. `newDiagSink` writes to `~/.local/state/omac/sandbox.log` whenever stderr is a terminal, so the VPN-down case the hint exists for was invisible to the agent that made the request. The client saw only `upstream dial failed: … connection refused`.

Then three smaller ones: `prompt:needs_intent` returned early and carried no hint; the DNS-failure body claimed `DENIED BY THE SANDBOX network policy` and listed three policy knobs while the hint below it said nothing had been denied by policy; and hard-deny bodies offered `allow_domain`, which provably cannot override them.

## How

- `registryhint.go` (renamed from `registry.go` — `registry` in this repo means `internal/registry`, the skill registry): remedy on its own line; `needs_intent` case added; `flattenForLog` keeps the log form single-line for the diag sink's per-line columns and whole-line appends.
- `server.go`: `upstreamFailureResponse(host, err)` puts the hint in the 502 body and de-duplicates a branch repeated verbatim across `handleConnect` and `handleForward`. `denyBody` switches on reason class — intent / declined intent / resolution failure / built-in guard / policy.
- Merge resolution with #157 (`denyBody(host, Verdict)` + intent state machine): both kept, intent branches first, reason classes after. The two features disagreed on one path — a declined declared intent says *do not retry*, the registry note said *re-run and choose Allow*. The note is withheld there and kept on `needs_intent`, where retrying after naming the package is the point.
- `sandbox-network/spec.md`: the requirement said every filtered denial must name the policy knobs, which is wrong for the two verdicts that are not policy decisions. Now scoped by class, with four scenarios.

## Verification

At `7348f77`:

- `go build ./...`, `go vet ./...`, `gofmt -l` clean; `go test ./...` green (30 packages); `go test -race ./internal/...` green.
- **omac-tng-installer e2e built from this SHA** (`E2E_OMAC_REF`): **52 PASS / 0 FAIL / 1 SKIP** (live layer opt-in). Includes `@tngtech/opencode-skainet installs in-sandbox (skainet provider registered, cold cache)` — the cold-cache plugin install this feature is about. An earlier run showed 1 FAIL on `internal registry reachable in-sandbox`; that was the host being off VPN (host probe reported `HTTP 000` too), and it passes with the VPN connected.
- New tests assert on bytes read off a real proxy connection: hint present in the 502 body, no body line over 100 chars, deny hint within that width across all eight reason branches, logged form single-line, and `allow_domain` cannot rescue a hard-denied host. Real policy denials are held byte-identical by a counterweight test over all five policy reasons.
- Pre-existing `TestServer502OnUpstreamFailure` / `TestServer502OnUpstreamDialFailure` pass unchanged. The e2e net-denial assertions (`security_assertions.go:104`, `provenance_test.go:218`) are unaffected — OR-lists that still match on `403`.

## Follow-up

- CI never created a `pull_request` run for #148's head `8fbb76d` (12 min, no run) though CI is healthy repo-wide; this PR should produce one, including the macOS leg.
- `scripts/e2e-docker.sh cache` fix is unrelated to the feature and could be split out if preferred.


nhuelstng:
## Verification complete — live layer included

Ran the remaining suites at `7348f77`.

### omac-tng-installer e2e, built from this SHA, `E2E_LIVE=1`

**57 PASS / 0 FAIL / 0 SKIP.** All five live completions land:

```
PASS claude as-installed (install.sh config -> api.anthropic.com)
PASS claude subscription completion (intended)
PASS claude via taia (install.sh CLAUDE_PROVIDER=taia config)
PASS opencode skainet completion (cold per-workdir cache, plugin installed in-sandbox)
PASS codex skainet completion
```

The opencode one is the path this PR is about: real model traffic through a cold per-workdir cache with the `@tngtech` plugin installed in-sandbox from the internal registry.

### omac e2e (Docker)

| Tier | Result |
|------|--------|
| cache | 9 tests, 6 subtests, **0 skips** (go, go_module_proxy, npm, pip, rustc, cargo) |
| echo lifecycle (opencode) | PASS, 130.85s |
| security audit (opencode) | 11 properties pass, `fsReadDenied` fails — **pre-existing, see #180** |

`fsReadDenied` is not a regression here: `origin/main` (`b9419da`) fails identically, same property, same eleven passes. It is a false positive under root — `/etc/shadow` is DAC-protected rather than masked, and the audit runs as uid 0 in a `--privileged` container. Verified by running the same test binary under the real sandbox at two uids (readable at uid 0, `Permission denied` at uid 1001). Filed as #180 with the evidence; out of scope for this PR.

Network isolation passes in that audit, which is the property this PR could plausibly have broken — the deny-body rewrite keeps the marker the audit greps for.

### Full matrix

| Suite | Result |
|-------|--------|
| `go build` / `vet` / `gofmt` / `go test ./...` | clean, 30 packages |
| `go test -race ./internal/...` | clean |
| CI (Linux + macOS) | 10/10 pass |
| omac e2e cache / echo | green |
| omac e2e audit | 11/12, the 12th pre-existing (#180) |
| installer e2e @ this SHA, live | 57/57 |

PR #188, opened 2026-07-29
Fix/Merge briefing into single system block

### LINKED ISSUE #171: Briefing Injection conflicts with some models

tmertztng:
**Issue:** Closes #171 

## What
* merge briefing into the existing system block instead of adding a second one.

## Why
* Some models reject a system message at index > 0 ("system message must come first").

## Verification
| Command | Result |
| ---- | ---- |
| gofmt -l . && go vet ./... && go vet ./... -tags=e2e && go vet ./... -tags=e2e_fast &&  go build ./...	| clean |
| go test -race -count=1 -timeout=5m ./...	| exit 0 · 29 packages · no races |
| go test -tags=e2e_fast ./internal/e2e/	| ok |
| go test -tags=e2e -run 'TestHarnessCLIContract\|TestHarnessLaunchProbe' ./internal/e2e/	| ok, 45.8s — existing callers of the changed helper |
| go test -tags=e2e -run TestOpenCodeSingleSystemMessage ./internal/e2e/	| PASS, 10.1s — 2 requests, 1 system message each |
| scripts/e2e-local.sh smoke opencode	| PASS — contract 5/5 + launch probe, opencode 1.17.12 |

### DIFF
--- a/.github/workflows/e2e-smoke.yml
+++ b/.github/workflows/e2e-smoke.yml
@@ -145,14 +145,14 @@ jobs:
             echo "version=main@$(git describe --tags --always --dirty)" >> "$GITHUB_OUTPUT"
           fi
 
-      - name: Model-free smoke (contract + launch + serve)
+      - name: Model-free smoke (contract + launch + serve + system prompt)
         id: modelfree
         env:
           E2E_CACHE_DIR: ${{ runner.temp }}/e2e-install-cache
         run: |
           set -o pipefail
           go test -tags=e2e -timeout=20m -v \
-            -run 'TestHarnessCLIContract|TestHarnessLaunchProbe|TestHarnessServeProbe' \
+            -run 'TestHarnessCLIContract|TestHarnessLaunchProbe|TestHarnessServeProbe|TestOpenCodeSingleSystemMessage' \
             ./internal/e2e/ 2>&1 | tee modelfree.log || true
 
       - name: LLM turn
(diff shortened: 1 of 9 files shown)


NoRiceToday:
### Scope: does this fully close #171?

#171 is harness-agnostic ("Qwen expects system prompt first"), but this PR only fixes the OpenCode path. The pi bridge still does the literal anti-pattern this PR removes:

- `.pi/extensions/omac-bridge.ts:165-167` — `messages.unshift({ role: "system", content })` injects a second system message at index 0. A pi user on a Qwen-like provider still hits "system message must come first" after this merges.
- `TestBriefingMergesIntoExistingSystemBlock` guards only the OpenCode plugin; the pi bridge's `unshift` is unguarded.
- The only wire-level e2e is OpenCode-only; pi (the vulnerable harness sharing the bug class) has none.

Two options: (a) extend the same merge-into-last-system-block pattern to the pi bridge + add an analogous guard/test, and close #171 here; or (b) merge this as the OpenCode fix and edit #171 to note it's partially addressed, with a follow-up for pi.

### Nits

- The test asserts the briefing *arrived* (`"omac sandbox"` substring) but not that it's *intact* / correctly ordered before the manifest. A stronger assertion on ordering/content would lock the invariant.
- Two byte-identical copies of the plugin still exist and both were edited. The drift guard is a test that silently skips when the canonical copy is missing, so a future divergence wouldn't fail CI — worth tightening or consolidating.

Issue #119, opened 2026-07-17
Network prompt unusable for proxy-unaware clients (JVM/gradle/maven): no popup, silent Landlock block; proxy-configured → 407; shown dialogs self-close


nhuelstng:
## What

The built-in sandbox network prompt is **unusable for proxy-unaware clients**
(JVM/gradle/maven, and similar tools). Running a Gradle build (or any JVM
dependency download) inside the omac sandbox fails to reach the network, and in
the common cases **no permission dialog ever appears** — the request is silently
killed by the kernel before the filter runs. In the cases where a dialog *does*
appear, it self-closes before it can realistically be answered.

Concretely, three distinct gaps:

1. **Default gradle never reaches the proxy → no popup, silent block.** The JVM
   ignores `HTTP_PROXY`/`HTTPS_PROXY` env (`ProxySelector` returns `[DIRECT]`),
   so it opens a direct connection, which Landlock blocks with
   `java.net.BindException: Permission denied`. The proxy filter is never
   consulted, so no prompt can fire.
2. **Even a proxy-configured JVM fails auth → 407, still no popup.** Pointing the
   JVM at omac's proxy via `-Dhttps.proxyHost/Port` reaches the proxy, but the
   JVM does not forward the session token (it lives in the `HTTP_PROXY` userinfo,
   and the JVM doesn't send `Proxy-Authorization` for CONNECT). omac returns
   `407 missing/invalid proxy token`; the filter/prompt is never reached.
3. **When a dialog does appear, it is not reliably answerable.** For a
   proxy-aware + token-authed client the dialog fires, but it is not tied to
   being answered — it closes on its own when the client times out or the session
   tears down (observed lifetimes 2.7 s / 11 s / 33 s across runs). Once the
   triggering connection is gone, clicking "Allow" achieves nothing.

## Why

- Real projects resolve dependencies with Gradle/Maven; the agent must run these
  unattended inside the sandbox.
- Today the outcome is a confusing hard failure (`BindException: Permission
  denied`) with **no feedback and no way to grant access** — the user cannot tell
  it is a sandbox network-policy denial, and the prompt that is supposed to let
  them allow it either never shows or vanishes too fast.
- Allowlisting the host does **not** help: Landlock blocks the JVM's *direct*
  connection regardless of `allow_domain`; only proxy-routed + token-authed
  traffic is ever filtered.

## How (root cause)

[...]

Issue #92, opened 2026-07-15
Authenticated package-registry access without exposing the token to the agent


nhuelstng:
# Authenticated package-registry access without exposing the token to the agent

**Labels:** enhancement, security

## What

Let a sandboxed agent install dependencies from **private, authenticated
package registries** (npm, Maven first; pip/Go/Cargo/NuGet next) **without the
registry credential ever entering the sandbox**.

Do it with **one generic mechanism**, not a per-tool solution: a host-side
authenticating mirror plus a small **declarative registry catalog** in which
the built-in tool types are just presets over a user-editable schema. A user
whose service matches a supported schema can add it in config — no omac
release required.

## Why

- Real projects resolve dependencies from private registries that require a
  bearer/basic credential; the agent must run `npm install` / `mvn` unattended.
- Handing the agent the token is a non-starter — omac's model is that secrets
  live in the OS keychain and are **never reachable inside the sandbox**.
- The obvious workarounds are wrong:
  - **Skill/sidecar** serves *explicit runtime API calls*; dependency
    resolution is implicit build infra with nothing to expose over the facade.
  - **Config-file injection** (`.npmrc`, `settings.xml`) puts the token on a
    path the agent can read.
- omac is moving **off nono**, so nono's `custom_credentials` /
  `--network-profile` path is not a solution to depend on. This must be native.

## Current gap

omac's native proxy (`internal/netproxy`) deliberately **never terminates
TLS** — CONNECT is a raw byte tunnel. It cannot see or add headers to HTTPS
registry traffic, so there is no native credential-injection path today.

## Proposed solution — generic authenticating mirror

A host-side (unsandboxed, in the supervisor) reverse proxy the sandboxed
package manager talks to over loopback. It forwards to the upstream registry
and injects the credential on the way out. The agent only ever speaks plain
HTTP to `127.0.0.1`; the token stays in the supervisor process (from the
keychain) and never crosses into the sandbox.

```
  ┌──────── sandbox ────────┐        ┌──── supervisor (host) ────┐
  │ npm / mvn               │        │ registry-mirror           │
  │  registry=127.0.0.1:P ──┼──open──┼─▶ inject Authorization     │
  │  (no token in config)   │  port  │   real TLS ──▶ upstream    │
  │                         │◀───────┼── stream response          │
  └─────────────────────────┘        └────────────────────────────┘
     plain HTTP, localhost only         authenticated TLS, host→upstream
```

### No committed-file changes

[...]

PR #192, opened 2026-08-04
feat(build): add mediated JVM build executor

### LINKED ISSUE #191: stableport: scanned neighbor counts as "fallback" and is never persisted — permanent fallback-ephemeral warn-loop
# stableport: scanned neighbor counts as "fallback" and is never persisted → permanent fallback-ephemeral warn-loop

**Duplicate check:** searched open issues (stableport / fallback ephemeral) — no dup.

## Context / Summary

The container proxy (and credproxy, same `stableport.Select` pattern) logs `using fallback ephemeral port <P> (stable window unavailable; warm-daemon DOCKER_HOST may drift on next run)` on **every** `omac build` invocation for an affected worktree, even though the hashed preferred port and its neighbors are free. Observed live on yarp3 (macOS, worktree bound to `containerproxy-port` = 38087).

## Problem / What

Reproduced sequence (all steps verified on the host, 2026-08-03):

1. `<cache>/gradle/.omac-control/containerproxy-port` contains `38087` (persisted correctly on an earlier run).
2. `lsof -nP -iTCP:38087` → empty; a bare `nc -l 38087` on the host succeeds; and an independent reimplementation of `IsFree`/`Select` in a throwaway Go program reports **38087 free**. Port occupancy is excluded — nothing listens and the port binds fine outside the omac process.
3. Nevertheless every build binds **38088** (preferred+1) and logs the fallback warning. 38088 = `preferred + 1`, i.e. `Select` returned a scan-window neighbor, meaning the `IsFree(38087)` probe failed **from inside the omac build process only**.

Two compounding code problems:

1. **`choosePort` treats any `chosen != preferred` as "fallback" and skips persistence** (`internal/containerproxy/proxy.go:382-406`, `return chosen, chosen != preferred`; persistence gated on `!fallback`, `proxy.go:333-365`; same shape in `internal/credproxy/proxy.go:~296`). A scanned neighbor is a perfectly good stable port, but because it is never written back, the next run re-reads the stale preferred (38087), fails `IsFree` again for the same (still unknown) reason, scans to a **new** neighbor, and warns again — a permanent warn-loop with the noisy, scary message pinned on every build.

2. **`IsFree` has no distinguishable failure reason** (`internal/stableport/stableport.go:57-64`): it returns plain `false`, so the log cannot say *why* 38087 was unbindable from omac's process (EPERM vs EADDRINUSE vs sandbox context). The externally-verified fact that the port binds fine from `nc` and from a standalone Go binary makes this look like a process-environment-dependent bind failure that the current code cannot observe or report.

**Second, independent observation** (candidate for its own ticket): a **warm Gradle daemon outlived its cred-lift proxy** and kept dialing the cached mirror URL (`GET /id-gitlabcom-yarp3/...` on `127.0.0.1:38087` — captured by `nc -l` in place of the dead listener). Denials surfaced as `net DENY gitlab.com:443 (missing/invalid proxy token)` — misattributed noise against the *filtered* proxy, because the daemon also probed the declared upstream directly. The stale warm daemon silently probes dead ports; omac does not detect or recycle it. (The `omac build stop` recycle did clear the daemon; the port warning persisted anyway, confirming problem (1) is independent of warm daemons.)

## Suggested fix / Ask

- [ ] `choosePort`/`Select`: persist any selected port that lies **inside [StablePortMin, StablePortMax)** (preferred *or* scanned neighbor) via `WritePreferred`; skip persistence only for a true out-of-window random/ephemeral fallback. This breaks the warn-loop: next run prefers exactly what this run bound.
- [ ] `IsFree`: return/record the bind error (at least under verbose logging) so "stable window unavailable" can be diagnosed — e.g. `window scan: preferred P: listen tcp 127.0.0.1:P: <errno>` instead of a bare true/false.
- [ ] (Separate ticket, optional) warm-daemon stale-URL self-heal: at build start, compare the daemon's cached cred-mirror/DOCKER_HOST port with the port this run actually bound; on mismatch, recycle the daemon (`gradle --stop` on the leaf) once instead of letting it probe dead loopback ports.

## Non-goals

- No change to the deny-by-default filtered-proxy allowlist (the `net DENY gitlab.com (not in allowlist)` lines are intended behavior — private fetches must route through the credential-lift proxy).
- No change to stable-port hashing (`stableport.For`) or the [30000,40000) window.
- Not a yarp3 environment fix: the local workaround is `omac build stop --root backend` + deleting stale daemons; this ticket is about the product-level warn-loop and diagnosability.

## Evidence / Environment

- Host: macOS 26.x (aarch64), Mullvad VPN installed (pf potentially in play); yarp3 worktree `/Users/sajjadtng/Documents/Ippen/ippen-digital/core/yarp3`.
- Control file: `~/.cache/omac/<hash>/gradle/.omac-control/containerproxy-port` = `38087` (mtime Aug 3 08:16); `credproxy-port` present too.
- Every build log line: `omac build: containerproxy: containerproxy: using fallback ephemeral port 38088 (stable window unavailable; warm-daemon DOCKER_HOST may drift on next run)` — even after `pkill -f GradleDaemon`, `omac build stop`, and `lsof` confirming 38087 free.
- Independent probe (Go, FNV-1a of the canonical worktree path, same Select/IsFree logic): `preferred=38087; first free=38087` — the port binds fine outside the omac process.
- Builds themselves succeed (`gradle :help`, `compileJava compileTestJava` BUILD SUCCESSFUL); impact is noise + real risk of warm-daemon DOCKER_HOST drift for Testcontainers lanes.

### LINKED ISSUE #207: Build brokered canary: e2e test for the full omac build loop (Gradle + Testcontainers)
## Context / Summary

`internal/cli/build_integration_test.go:37-43` carries an explicit TODO: no test exercises the full JVM-build loop end-to-end (`omac start` sandbox → brokered `omac build` → host build broker → restricted executor → a real Gradle wrapper + tests). The closest coverage is unit/integration: stub `gradlew` shells, engine-contract tests, and broker/authorizer tests. The first place the full loop would break is today a client's real repository — i.e. after we've shipped the feature, not before.

This issue adds a **build brokered canary** (model-free, deterministic) that runs that loop in CI so a JVM-build regression blows here, not on client work. Term added to `CONTEXT.md`; see ADR 0001 (executor scope), 0002 (mediated container access), 0003 (macOS Shape A), 0004 (in-process broker) for the load-bearing decisions.

## Problem / What

- `internal/cli/build_integration_test.go:37-43` — comment marks host-side validation against a real Gradle wrapper as TODO; assertions use stub wrappers only.
- No test asserts a **cold Gradle daemon per build through the brokered path** — the property ADR 0001's 2026-08-04 revision made load-bearing (spring-kafka `GlobalEmbeddedKafkaTestExecutionListener` staleness on warm daemons, init-script closure caching at daemon-config time).
- No test asserts **mediated container access** (ADR 0002) works end-to-end: `.omac/build.yaml` image allowlist → container proxy bind → `DOCKER_HOST` injection → Testcontainers → executor-owned cleanup. The container proxy and its label filter are today covered only by unit/fake-upstream tests.
- Approval state flow is covered piecemeal (`build_approve_test.go`, `serve_snapshot_test.go`) but no test drives the brokered path with a pre-approved manifest inside a managed session.

## Suggested fix / Ask

Add `TestE2EJvmBuild` — one test function in `internal/e2e/` (build tag `e2e`, model-free — no `SKAINET_*` secrets needed), driving the loop exactly as an agent does, with assertions on the filesystem artifacts Gradle writes. Sub-items:

- [ ] **Fixture** `internal/e2e/testdata/jvm-fixture/` — a minimal synthetic Gradle project (wrapper, Groovy DSL, JUnit 5), deliberately NOT a copy of any client repo:
  - One unit test `GreetingServiceTest` using a Mockito-mocked `GreetingRepository` — covers the plain unit-test path through the executor.
  - One IT-named `PostgresIT` using Testcontainers `PostgreSQLContainer` (`postgres:16-alpine`) with one `SELECT 1` JDBC round-trip — covers the container-proxy path (ADR 0002).
  - **No Spring.** The property being canaried is cold-daemon-per-build (listener re-registration + `init.d/` re-read), not any framework-specific class. Spring/Kafka would multiply fixture weight without asserting a distinct omac property.
  - Committed `.omac/build.yaml` declaring the `postgres:16-alpine` image.
- [ ] **Approval pre-seed (no TTY)** — host-side test setup imports `internal/buildmanifest`+`internal/buildcontrol`, computes `digest`/`caps` for the fixture manifest, and writes `<cacheRoot>/build-control/approvals/<sha256(canon-worktree)>.json` via `ApproveAt` BEFORE launching the parent — the `internal/cli/serve_snapshot_test.go:56-66` recipe. This sidesteps the TTY-only `omac build approve` (`internal/cli/build_approve.go:189-200`) without any production-code change or auto-approve escape hatch (per `docs/build-command.md:229`, none exists and none should be added).
- [ ] **Invocation** — reuse `runOmacShellWithEnv` (`internal/e2e/fixtures.go:748`) to launch `omac start claude-code --inner /bin/sh -- omac build --root jvm-fixture -- gradle test`. Harness-locked to `claude-code`: the broker is harness-agnostic (ADR 0004); the harness exists here only for the `--inner` seam and its binary is never launched into a model.
- [ ] **Assertions** — on the Gradle-written XML reports in the fixture's `build/test-results/test/`:
  - exit code 0
  - both `TEST-…​GreetingServiceTest.xml` and `TEST-…​PostgresIT.xml` exist with `failures="0" errors="0"`
  - the executor-owned container/network from the session is gone after teardown (label-filtered `docker ps -a --filter label=…` empty), proving the proxy's lifecycle cleanup (ADR 0002) not just its data path.
- [ ] **Cold-cache pre-seed with loud failure** — probe `<cacheScope>/gradle/wrapper/dists` for the Gradle dist; if absent, run a host-side `./gradlew --version` in the fixture ONCE to seed; if it still cannot be made warm, `t.Fatalf` (mirrors AGENTS.md's cache-isolation "missing toolchain = broken image, not a skip" rule).
- [ ] **Nested-sandbox branch** — when `nestedInOmacSandbox()` is true (`scripts/e2e-local.sh` drives with `--no-sandbox`), run unit-test leg only and record document-exposure for the IT leg, mirroring `sandboxActive := !forceNoSandbox(h)` semantics in the audit test.
- [ ] **Local iteration tier** — add a `build` tier to `scripts/e2e-local.sh` mapped to `-run '^TestE2EJvmBuild'`, with the `E2E_NESTED`/`E2E_RECOVER_INSTALL`/`TMPDIR` accommodations already provided for other tiers.
- [ ] **CI workflow** `.github/workflows/e2e-build.yml` — separate from `e2e.yml` (keeps the canary isolated from the model-flaky matrix), `on: pull_request` + `workflow_dispatch`:
  - **Unit leg**: `macos-latest` — no Colima, runs the test with the IT leg skipped via env.
  - **IT leg**: `macos-15-intel` — 14 GB Intel (the ONLY GHA tier where a Linux VM can run: arm runners have nested-virt disabled; verified via docs + `setup-docker-macos-action` support matrix). Steps: `brew install colima docker`, `colima start --memory 6 --cpu 2`, export `DOCKER_HOST=unix://$HOME/.colima/default/docker.sock`, `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock`, `TESTCONTAINERS_HOST_OVERRIDE=$(colima ls -j | jq -r .address)`, then run the full test.

## Acceptance criteria

- [ ] `go test -tags=e2e -run '^TestE2EJvmBuild' ./internal/e2e/` passes locally on macOS with Colima running.
- [ ] Same command passes in a nested omac sandbox via `scripts/e2e-local.sh build` with the IT leg in document-exposure mode.
- [ ] The new `e2e-build.yml` is green on PR: unit leg on `macos-latest`, IT leg on `macos-15-intel`.
- [ ] Breaking the daemon-recycle (e.g. reverting `gradlew --stop` post-build) or the image allowlist produces a LOUD failure of this test — specifically not a skip.
- [ ] ~~Removing the pre-seeded approval file produces exit 3 with the "manifest approval required" diagnostic, and this negative case is asserted as a sub-test.~~
- [ ] Removing the pre-seeded approval file produces **exit 10** (service failure) with the "no parent capability snapshot for this worktree" diagnostic AND the "omac build approve" + restart hint, asserted as a sub-test. The brokered path is a service failure by design: the parent must restart to re-freeze the in-memory snapshot (ticket-06 freeze-once), so the build is not retryable in the same session.
- [ ] No `.github/workflows/e2e.yml` change; no production-code change outside `internal/e2e/` + `scripts/e2e-local.sh` + the new workflow file.

## Non-goals

- **No changes to `internal/cli/` / `internal/buildbroker/` / `internal/containerproxy/` / `internal/buildengine/`** — this is a test-only canary. If the canary exposes a broker bug, that becomes its own issue.
- **Not** a lift of the darwin-only container-proxy gate (`internal/cli/build_proxy.go:206`) to Linux. That is a feature change with its own design discussion; the IT leg tests the feature where it exists today (macOS).
- **Not** a client-repo fixture (no yarp3/backend checkout). The fixture is synthetic and committed, so the canary is deterministic and requires no external repo or secrets.
- **Not** Spring/Kafka replication of the original ADR 0001-revision failure mode. Cold-daemon-per-build is the property under test; the spring-kafka `GlobalEmbeddedKafkaTestExecutionListener` staleness was the motivation, not the asserted property.
- **Not** a harness matrix. One harness (claude-code) for the `--inner` seam; the broker path is harness-agnostic.
- **Not** adding a TTY-less approve escape hatch (`OMAC_BUILD_AUTO_APPROVE`-style) — that would weaken the security property the approval gate exists for. The pre-seed mechanism uses the existing `ApproveAt` API the repo already uses in `serve_snapshot_test.go`.
- **Not** Testcontainers Cloud or any SaaS dependency, and **not** self-hosted runner hardware.

## Evidence / Environment

- TODO comment: `internal/cli/build_integration_test.go:37-43`
- Broker env wiring: `internal/cli/build_broker_wiring.go:192`, mounted by `internal/cli/start.go:853-922` / `serve.go:355+`
- TTY check that rules out scripted approve: `internal/cli/build_approve.go:189-200`
- Pre-seed recipe already used in-repo: `internal/cli/serve_snapshot_test.go:56-66`
- Approval file location schema: `internal/buildcontrol/buildcontrol.go:137-140, 152-154`; `internal/buildmanifest/approval.go:381-394`
- macOS-only container proxy gate: `internal/cli/build_proxy.go:205-210`
- GHA macOS nested-virt constraint: <https://docs.github.com/en/actions/reference/runners/github-hosted-runners> (arm64 runners: "Nested-virtualization is not supported"); actionable Docker-on-macOS GHA path confirmed Intel-only by <https://github.com/douglascamata/setup-docker-macos-action>
- Colima + Testcontainers config: <https://java.testcontainers.org/supported_docker_environment/>
- Domain terms: `CONTEXT.md` in the artifact workspace (build brokered canary; build proxy triad; mediated container access; executor session; build control state)

NoRiceToday:
**Issue:** Closes #191 · Closes #207 (build brokered canary) · Refs #92 · Implements the host build broker follow-ups (spec: `.scratch/jvm-build-executor-follow-ups/spec.md`, gates 1–7)

## What

- Add `omac build` for constrained Gradle requests with protected control state, isolated cache, JDK resolution, cancellation, and leaf-keyed serialization.
- Let projects declare non-secret build capabilities in `.omac/build.yaml`, subject to host approval and policy ceilings.
- Add host-side credential and Docker proxies so builds can resolve approved private Maven dependencies and use approved Testcontainers images without receiving credentials or the raw daemon socket.
- Add the host build broker (`internal/buildbroker`): a session-scoped, in-process broker in the unsandboxed `omac start`/`serve` parent that accepts constrained build requests from a sandboxed `omac build` thin client over the loopback control plane. The broker owns host-only orchestration (manifest/policy evaluation, keychain access, proxy lifecycle, leaf-keyed serialization, restricted executor launch, cancellation, cleanup) while repository-controlled build code still runs only inside the restricted JVM build executor (ADRs 0001, 0002, 0004).
- Add the managed-mode marker (`OMAC_BUILD_BROKER_REQUIRED=1`) with fail-closed behavior, the per-parent cryptographically random build token, and `omac build approve` (host-only, interactive-terminal-only) plus the parent-restart requirement for capability activation.
- Add the pending-to-active daemon ownership handshake, platform-tested process identities (`internal/procidentity`), parent-startup reconciliation, and verified trusted daemon control for brokered `omac build stop` (never executing the repo wrapper with host authority, never removing the lockfile).
- Add the model-free build brokered canary `TestE2EJvmBuild` (issue #207): a committed synthetic Gradle fixture (JUnit 5 + Mockito + Testcontainers, no Spring) driving the full brokered loop through a real Gradle wrapper, with approval pre-seed, cold-cache loud failure, an approval-gate negative subtest, an IT leg asserting executor-owned container/network cleanup, and a dedicated `e2e-build.yml` CI workflow (unit leg `macos-latest`, IT leg `macos-15-intel`).
- Correct stale comments and docs: per-worktree lock claims → leaf-keyed serialization with the lock in host-only `build-control/`; max-duration expiry described as immediate force → graceful-then-forced staged path; `docs/build-command.md` now describes the managed build path alongside the unchanged direct host path.
- Document the supported workflow and the macOS/Linux network-boundary difference.

## Why

Direct Gradle execution in the agent sandbox is unreliable for daemon, worker, and Testcontainers workflows (macOS keychain access is denied by the outer Seatbelt profile; the Colima socket is unavailable to the container proxy; a nested Gradle Seatbelt profile is rejected). Running it unsandboxed would execute repository-controlled build code with host-user authority.

`omac build` keeps that code in a restricted executor while supplying only the capabilities a declared build needs. The broker moves host-only orchestration (keychain, raw daemon, proxies) into the unsandboxed parent so the agent sandbox never receives that authority, while the build executor stays restricted.

## How

```mermaid
flowchart LR
    H[Harness] -->|omac build| G[Manifest and host-policy gate]
    G -->|approved| E[Restricted Gradle executor]
    G -->|denied| D[Actionable policy denial]

    E --> C[Protected control state<br/>queue, cache, JDK, temp]
    E -->|public dependencies| P[Filtered network proxy]
    E -->|private Maven| R[Credential-lift proxy]
    R --> K[Host keychain]
    E -->|approved images| X[Docker policy proxy]
    X --> Docker[Host Docker daemon]
```

- The executor receives no ambient Gradle home, host secrets, raw Docker socket, or unrelated host paths.
- The Docker proxy is allowlist-based, injects executor ownership, and rejects privileged or host-escaping container options.
- Stable proxy ports and post-build daemon recycling prevent stale daemon connections across runs.
- Host build mediation runs in-process in the unsandboxed session parent; a clean broker/build-engine seam keeps a future managed-sidecar move straightforward without adding sidecar lifecycle now (YAGNI/KISS, ADR 0004).
- Proxy-mediated dependency, credential, and container support is macOS-only in v1 (Shape A env-only filtering); on Linux the executor is kernel-blocked and those loopback proxies are not started — Linux proxy support is deferred.
- Brokered execution is fail-closed: the parent injects `OMAC_BUILD_BROKER_REQUIRED=1` even on setup failure; a partial broker tuple exits 10 with a restart/upgrade diagnostic; managed invocation never falls back to nested local execution.
- Brokered `omac build stop` uses verified daemon control (`procidentity.Verify` + host-only ownership records) — it never executes the repo wrapper, never applies a relaxed profile, never removes the lockfile.

### Broker work gates (spec `.scratch/jvm-build-executor-follow-ups/spec.md`)

1. Create and self-assign the focused issue.
2. Extract `internal/buildengine` without changing ordering, exit behavior, or existing direct invocation (ticket 04).
3. Add `internal/buildbroker`, the finite execute protocol, cancellation endpoint, explicit managed-mode marker, and `start`/`serve` wiring (ticket 05).
4. Move lock acquisition before mutable control state; correct the lock's cache-leaf semantics; relocate the lock to host-only state; namespace worktree-specific records; freeze active capabilities in parent memory; add the host-only approval transition; enforce cross-platform read-only control-state projection; remove lockfile deletion (ticket 06).
5. Keep post-build recycle inside the restricted executor lifecycle and replace brokered manual wrapper stop with the pending-to-active daemon handshake, platform-tested process identities, startup reconciliation, and verified trusted daemon control (ticket 07).
6. Enable brokered build and stop only after all security and lifecycle gates pass on macOS and Linux (wired in ticket 07 Phase 5; e2e validation deferred to the `e2e.yml` macOS/Linux matrix).
7. Update `docs/build-command.md`, code comments that claim per-worktree locks or immediate max-duration force, and PR #192's issue link and verification (ticket 08 — this commit).

## Verification

- `go vet -buildvcs=false ./...` passed.
- `go build -buildvcs=false ./...` passed.
- `go test -count=1 ./internal/buildrun ./internal/buildengine ./internal/buildbroker ./internal/buildcontrol ./internal/procidentity` passed.
- `go test -count=1 ./internal/cli` (excluding the pre-existing sandbox-only `TestDoctorHarnessBinarySection`) passed.
- `git diff --check main...HEAD` passed.
- Full `go test ./...` remains blocked in this nested sandbox: the doctor test cannot read `~/.config/omac/config.yaml`, and macOS integration tests cannot apply a nested Seatbelt profile. These run on host/CI.
- Ticket 08 is prose and comments only — no behavior changes. The affected packages' tests are green unchanged.

## Follow-up

- #92 remains open: this delivers the Gradle/Maven credential-lift tracer bullet, not the issue's generic multi-tool registry catalog.
- Real-host validation remains for canonical worker checks, cold container compilation, and the multi-harness JVM build workflow.
- E2E validation of the daemon ownership handshake + brokered build/stop (host-gated macOS keychain + fake Docker endpoint; Linux bwrap-from-parent + private-loopback) is deferred to `internal/e2e/daemon_ownership_test.go` driven by `scripts/e2e-docker.sh` and the `e2e.yml` macOS/Linux matrix (gate 6).

## Reusable components

The executor introduces four components intended for reuse beyond the Gradle path:

- **Build capability approval** (`internal/buildmanifest`): a project-declared `.omac/build.yaml` capability contract vetted by host policy against a frozen-for-session gate — a seam for future build tools (e.g. a Maven adapter) rather than Gradle-only logic.
- **Credential lift** (`internal/credproxy`): a host-side loopback proxy that injects keychain `Authorization` upstream for approved registries without the credential ever entering the executor — a partial step toward the generic authenticating mirror in #92, usable beyond Maven.
- **Docker mediation** (`internal/containerproxy`): an allowlisted, executor-owned Docker-compatible endpoint with scoped prune and denial correlation — reusable by any sandboxed workflow needing mediated container lifecycle without raw daemon access.
- **Stable-port components** (`internal/stableport`): per-worktree loopback ports with a persisted control file and scan/ephemeral fallback — reusable by any host-side loopback service that must stay reachable across runs.

Related JVM proxy-routing work for proxy-unaware clients is tracked in #119 but not addressed in this PR.


NoRiceToday:
Finally in a state where I can safely say, that the client setup works fully with omac: unit tests, IT, test containers spawned using gradle, it all works now. Just writing a E2E test to assert that. 

What remains is to review this independently from my agent reviews and properly test it on Linux @nhuelstng - which is not easy given how big the diff has become...

NoRiceToday:
TODOS from sync, verify:
- test whether file access in tests is limited as intended
- on linux network access should be fully controllable

nhuelstng:
Blocking (security boundary does not hold)

1. Docker proxy: case-variant JSON keys bypass the whole HostConfig policy → privileged container + host bind mount = host root. internal/containerproxy/policy.go:216-217,402 reads exact-case keys from a map[string]any and re-marshals the raw map; moby decodes case-insensitively. {"Image":"<approved>","hostconfig":{"Privileged":true,"Binds":["/:/host"]}} is forwarded with 201. Same trick spoofs labels/image/hostIp. Reproduced by the reviewer against a copy of the package. Fix: decode into moby's typed structs and re-serialize only the typed struct; reject any key that EqualFolds a known key without exact case.
2. Docker proxy: %66ilters= defeats the ownership filter on every prune/list endpoint (policy.go:621,686 does HasPrefix(kv, "filters=") on the raw query; the daemon unescapes and takes the first value) → volumes/prune?%66ilters={"all":["true"]} wipes the developer's unused volumes/images/networks. Reproduced. Fix: url.ParseQuery, delete all filters, set the ownership filter.
3. Credential-lift proxy: path traversal turns it into an authenticated GET oracle against the whole registry host. internal/credproxy/proxy.go:482-484 concatenates the client path onto the upstream base with no .. rejection or prefix check → GET /internal/../../api/security/apiKey returns the developer's Artifactory API key. Confirmed.
4. Credential-lift proxy has no client authentication and no Host check (zero token handling in the file; deterministic FNV port persisted 0644, stableport.go:42-52,244). Any local process, other sandbox session, or browser via DNS rebinding can use the developer's registry credential. Confirmed.
5. macOS executor runs under (allow network*) (internal/sandboxrun/sbpl.go:126-129, selected by internal/buildrun/grants.go:494-497); Seatbelt classifies AF_UNIX connect as network-outbound, so build/test code can DOCKER_HOST=unix://~/.colima/default/docker.sock and skip the proxy entirely. Needs confirmation on a Mac, but if it holds the container proxy is advisory, contradicting docs :642-650 "the executor NEVER sees the raw daemon socket".
6. Direct host path runs the repo's mutable gradlew --stop unsandboxed with full host authority, after every direct build and on omac build stop (internal/buildrun/stop.go:92-102; internal/buildengine/engine.go:694 literally says UNSANDBOXED). A build that rewrites its own gradlew escapes the executor. Confirmed. Docs :424-443 don't mention it.
7. Daemon-ownership handshake trusts a self-reported pid. internal/buildrun/daemon_handshake.go:342-360 takes pid from JSON written by sandboxed code; procidentity.Verify only proves "some Gradle daemon on this JDK", no SO_PEERCRED/LOCAL_PEERPID or ancestry check → brokered omac build stop can be steered to SIGKILL the developer's IDE Gradle daemon. Confirmed. And on Linux the JVM reports a namespace-local pid (bwrap.go:46 --unshare-pid) that the host-side verifier looks up in host /proc — the feature can't work there as written.

Should fix before merge

8. Client-project-specific logic and naming committed into the tool. An omac-generated init script that retires tasks matching a specific client's checkstyle*Sandbox convention (internal/buildrun/control.go:244-308), plus that client's name in omac provenance output (internal/cli/provenance.go:247), CLI help (build.go:256) and docs — 26 mentions. This is both a client-name leak into a public repo and product-inappropriate behaviour (silently disabling users' Gradle tasks by name pattern). Remove; if needed, make it a manifest-declared opt-in.
9. Every design rationale points at documents that aren't in the repo. ADRs 0001–0004, CONTEXT.md, and the .scratch/…/spec.md the PR body links are all gitignored (.gitignore:69). Committed docs cite "ADR 0003 Revision" etc. COLLABORATION.md requires decisions to live in docs//ADRs. Reviewers currently cannot read what the code claims to implement.
10. Serve authorizer canonicalization mismatch — internal/cli/serve.go:1894,1258 keys active dirs by filepath.Abs, internal/buildbroker/authorizer.go:96-100 EvalSymlinks the client path before isActive → every brokered build under a symlinked dir (macOS /tmp, /var) fails 403. buildbroker.ServeActiveDirs (the type the tests cover) is dead code. Confirmed.
11. Serve mode: /__omac__/activate is unauthenticated and agent-callable (serve.go:1908), and with no --root the authorizer allows any active dir → an agent can activate ~/other-repo and get its gradlew executed with rw over it. Bind the broker to host-configured roots.
12. Docker proxy: containers stay on the default bridge (proxy.go:1256-1275 only connects to the executor network, never sets NetworkMode), and NetworkingConfig is unchecked → unrestricted egress and join-any-host-network. Image approval is tag-agnostic (policy.go:535-561, ?tag= ignored). Executor ID = worktree basename (build_proxy.go:444-453) → same-named worktrees scavenge each other's live containers. Upstream is hard-coded to Colima (proxy.go:153-156).
13. Credproxy secondary: repo-controlled upstream + alias-only approval rendering (approval.go:316-317) lets a second repo redirect the internal credential to https://attacker/; http:// upstreams accepted (proxy.go:251); Set-Cookie/WWW-Authenticate passed through (proxy.go:598-607); default redirect following (proxy.go:507). TestStart_ScansWhenStablePortBusy never calls Start() (vacuous).
14. Linux executor network posture breaks Gradle itself: grants.go:479-484 → Landlock --enforce with no ports denies loopback bind/connect, so client↔daemon fails; docs :539-560 describe a network namespace that doesn't exist (bwrap.go:22 "NOT --unshare-net"). e2e-build.yml is macOS-only so nothing exercises Linux.
15. Manifest decode is not strict despite the doc claim (manifest.go:178 node.Decode without KnownFields): typos silently drop capabilities and the approved digest covers the typed projection, not the file. ceilingStillValid (session.go:535-546) treats a removed ceiling as valid, contrary to its own comment. .omac/build.yaml as FIFO hangs the parent at start.
16. Control-state lifecycle: RetireDaemonOwnership is deferred unconditionally so a daemon surviving --stop becomes unowned/unstoppable (engine.go:645,873-885); startup reconcile deletes pending records without the leaf lock, killing a concurrent parent's in-flight handshake (reconcile.go:905-909); managed-mode Ctrl-C cancels the request ctx and returns exit 10 instead of the documented exit 4 (build_managed.go:200-215,287-299); two shutdown TOCTOUs (broker.go:294-297/359, 379-384 never closes req.done).
17. Global changes justified by build-only needs: /usr/libexec added to the macOS read baseline for every session (sandboxprofile/baseline.go:94); ExpandExisting now skips un-stat-able paths with a notice instead of erroring (expand.go:73-86) — safe direction, but a silent behaviour change for all users, motivated by running tests nested.
18. Docs drift (concrete): ~/.gradle/~/.m2 are not in protectedCommon() despite build-command.md:36,506,860; :266-271 "fail-closed allowlist" is advisory on macOS; CONFIGURATION.md +181-193 Linux claim false; bounds.go:82-85 says execute checks Accept: — it doesn't; authorizer.go:21-23 claims constant-time compare, uses !=.

VictoriaRuckerbauer:
Mathias and I have now rebased the branch on the current main.
Do I understand correctly that you are specifically trying to solve issues with Gradle projects?
And also, for priorization: do we have any users that are currently blocked by this?

Concerning the documentation:
- Likely, `build_command.md` now duplicates things already mentioned in other docs. We have not cleaned this up in the rebase.
- In my opinion, we need to focus on keeping documentation concise and well-reviewed. I do not think that maintaining 1000 lines of `build_command.md` is reasonable (or likely to happen), so I have a strong preference towards shortening it significantly. I think that for every section, it should be clearly visible what the documentation wants to tell the reader and why it is relevant to them.

Concerning the problem that is solved here: Is this something that could also be solved with VMs or similar, in case there are some workarounds that do not require such massive PRs? I think that we have a major conflict here between keeping things simple and well-reviewed, and supporting as many options as possible, so we should at least discuss any other options, if available.

NoRiceToday:
I had discussed this with Mathieu Desponds and @nhuelstng before building: I became a maintainer of omac because it's my daily driver sandbox and we need gradle to test in our $client project. Without it, agent is mostly guessing. I also showed omac in an AI meeting at $client and the colleagues were interested, and some TNG colleagues also tried it out, but never started using it, because of the lack of gradle support and (hopefully all fixed) setup issues. So my assumption would be that at least a few colleagues would also use it with this feature. And I'm sure many client projects use gradle. However, it is still quite a lot of complexity and a clear drawback (due to mac seatbelt limitations) that come with this, so we should all be certain that we want to add and maintain this feature. 

Regarding Architecture: AFAIK the only better way to solve this is using a (micro)VM. However, if we introduce a VM 'connection' to omac, I would ask myself as a user (and maintainer):  Why bother with omac if I already have a VM then? At that point, wouldn't it be simpler to fully commit a (micro)VM sandbox solution - both from usage and maintainer perspective?

NoRiceToday:
Fully agree on the documentation point

VictoriaRuckerbauer:
One option to check: can we open ports after the test runner has started (when the port is defined)?
Other alternative that might work: can omac figure out the port that the runner wants and ask the user whether it can open it?
How did nono solve this problem?

mwtng:
Rebased onto main after the "tngtech -> TNG" change in the imports

PR #200, opened 2026-08-04
fix(ci,skills): stop python sidecars reverse-resolving their bind address

### LINKED ISSUE #199: ci(macos): the model-free leg spends ~600s in getfqdn during stub startup
### Duplicate check

## Context / Summary

`E2E: model-free (macos-latest)` in `ci.yml` has taken ~700s on every run
since the stub-gateway test landed (#183, extended in #194), against ~60s
for the ubuntu leg. It passes, so nothing went red — it just became the
sole critical path of every CI run. Localized on branch
`diag/probe-model-macos-timing` (run 30944071802).

## Problem / What

All of the difference is `HTTPServer.server_bind()`, which resolves the
bind address with `socket.getfqdn()` between `bind()` and `listen()`. On a
macOS runner the reverse lookup of `127.0.0.1` has no answer and times out
after **~35s**, once per server process.

`scripts/probe-model_test.sh:89` (`start_stub`) restarts the stub 17 times,
so the file pays it 17 times.

Per-layer measurement (`scripts/diag-loopback-timing.py`, 20 repeats):

| measurement | ubuntu mean | macOS mean |
|---|---|---|
| python_startup | 16.0 ms | 56.7 ms |
| curl_spawn | 5.3 ms | 34.1 ms |
| **getfqdn** | 0.3 ms | **7 006 ms** (min 0, max 35 028) |
| server_bind (warm) | 0.1 ms | 0.3 ms |
| **stub_cold_start** | 112.8 ms | **35 277.6 ms** |
| curl_request | 5.2 ms | 11.5 ms |
| urllib_request | 1.4 ms | 2.2 ms |

The `getfqdn` row is the mechanism: the first call in a process costs 35s
and every later call 0ms (the resolver caches it), which is why
`server_bind` measured warm looks free and why each fresh stub process pays
it again. Requests themselves are 11.5ms — never the problem.

Projected from those rows: 621s macOS / 22s ubuntu. Measured: **658s / 23s**.

The same pattern sits in four more places, all of which pay the same 35s on
macOS:

- `.opencode/skills/echo-rest/scripts/sidecar.py:186`
- `.opencode/skills/self-audit/scripts/sidecar.py:69`
- `internal/facade/integration_test.go:85` (inline python sidecar)
- `CREATING_A_SKILL.md:454` — the documented sidecar template every new
  skill is copied from, and its mirror in
  `internal/builtinskills/assets/omac-write-a-skill/references/`

## Suggested fix / Ask

- [ ] override `server_bind()` to set `server_name` from the bind address
      instead of reverse-resolving it, in the probe stub and both sidecars
- [ ] same in the facade integration test's inline sidecar
- [ ] fix the documented template (+ mirrored asset) so new skills do not
      inherit the stall
- [ ] confirm `E2E: model-free (macos-latest)` drops to ~25s

## Non-goals

- Reworking `start_stub` to reuse one stub process across cases — with the
  lookup gone a restart costs ~110ms, so there is nothing left to win.
- The 20s of deliberate retry sleeps in case 10 stay: they are what that
  case tests.

## Evidence / Environment

- Diag run: https://github.com/TNG/oh-my-agentic-coder/actions/runs/30944071802
- macOS runner: Darwin arm64 25.5.0, python 3.14.6, curl 8.7.1, bash 3.2.57,
  hostname `iad20-...local`
- Affected CI runs: 30940353922 (705s), 30936589612 (708s), 30933922360 (700s)

nhuelstng:
**Issue:** Closes #199

## What

- `server_bind()` override that takes `server_name` from the bind address
  instead of reverse-resolving it, in the probe-model stub gateway and both
  sidecar skills
- same override in the facade integration test's inline python sidecar
- same in the `CREATING_A_SKILL.md` sidecar template and its mirrored
  builtin-skill asset, so new skills stop inheriting the stall

## Why

`HTTPServer.server_bind()` resolves the bind address with `socket.getfqdn()`
between `bind()` and `listen()`. On a macOS runner that reverse lookup of
`127.0.0.1` has no answer and times out after **~35s**, once per server
process. `scripts/probe-model_test.sh` restarts its stub 17 times, so
`E2E: model-free (macos-latest)` has taken ~700s against ~60s on
ubuntu-latest on every run since #183 — green, but the sole critical path of
every CI run.

Per-layer measurement (see #199 for the full table): `getfqdn` costs 35 028 ms
on the first call in a macOS process and 0 ms afterwards, `stub_cold_start`
35 277 ms vs 112 ms on Linux, while an actual request is 11.5 ms. Projection
from those rows was 621s; the traced test measured 658s.

`server_name` is only used to fill in CGI variables, which none of these
servers serve.

## How

The same four lines everywhere the pattern appears:

```python
def server_bind(self):
    socketserver.TCPServer.server_bind(self)
    self.server_name, self.server_port = self.server_address[:2]
```

- `scripts/probe-model_test.sh` — `Stub(HTTPServer)`
- `.opencode/skills/{echo-rest,self-audit}/scripts/sidecar.py` — `Server(ThreadingHTTPServer)`
- `internal/facade/integration_test.go` — inline `Server(ThreadingHTTPServer)`
- `CREATING_A_SKILL.md` + `internal/builtinskills/assets/omac-write-a-skill/references/creating-a-skill.md`

## Verification

Local (Linux):

- `bash scripts/probe-model_test.sh` → `all probe-model.sh checks passed`, 22.6s
  (unchanged; the 20s is case 10's deliberate retry sleeps)
- both sidecars started and answered `/status` → `{"ok": true, "skill": ...}`
- `go test ./internal/facade/ ./internal/builtinskills/` → ok (the latter is the
  doc/asset mirror-sync test)
- `go vet ./internal/facade/`, `gofmt -l internal/facade/` → clean
- `python3 scripts/check-docs.py` → `[ok] README.md 360/400 lines; all doc links resolve`

macOS is verified by this PR's own CI: `E2E: model-free (macos-latest)` is
expected to drop from ~700s to ~25s, and `Test (macos-latest)` to lose 35s from
the facade integration test. Numbers posted below once the run lands.

## Follow-up

The instrumentation that localized this — a per-case TRACE in
`probe-model_test.sh` and `scripts/diag-loopback-timing.py` — is on
`diag/probe-model-macos-timing` and deliberately not part of this PR. Say if
the timing script is worth keeping in-tree for the next platform-specific
stall; otherwise that branch gets deleted.

### DIFF
--- a/.opencode/skills/echo-rest/scripts/sidecar.py
+++ b/.opencode/skills/echo-rest/scripts/sidecar.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 import hashlib
 import json
 import os
+import socketserver
 import sys
 import time
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
@@ -179,11 +180,25 @@ class Handler(BaseHTTPRequestHandler):
         )
 
 
+class Server(ThreadingHTTPServer):
+    """ThreadingHTTPServer without the reverse-DNS lookup in server_bind().
+
+    The base class resolves the bind address with socket.getfqdn() between
+    bind() and listen(). On macOS that lookup of 127.0.0.1 has no answer and
+    times out after ~35s, delaying every sidecar start by that much before the
+    facade can reach it. server_name only fills in CGI variables, unused here.
+    """
+
+    def server_bind(self):
+        socketserver.TCPServer.server_bind(self)
+        self.server_name, self.server_port = self.server_address[:2]
+
+
 def main() -> int:
     if PORT == 0:
         print("echo-rest: $SIDECAR_PORT not set", file=sys.stderr)
         return 2
-    srv = ThreadingHTTPServer(("127.0.0.1", PORT), Handler)
+    srv = Server(("127.0.0.1", PORT), Handler)
     print(
         f"[echo-rest] listening on 127.0.0.1:{PORT} skill={SKILL} "
         f"secret={fingerprint(SECRET)}",
--- a/.opencode/skills/self-audit/scripts/sidecar.py
+++ b/.opencode/skills/self-audit/scripts/sidecar.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 import hashlib
 import json
 import os
+import socketserver
 import sys
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from urllib.parse import urlparse
@@ -62,11 +63,25 @@ class Handler(BaseHTTPRequestHandler):
         self._json(404, {"error": "not found", "path": self.path})
 
 
+class Server(ThreadingHTTPServer):
+    """ThreadingHTTPServer without the reverse-DNS lookup in server_bind().
+
+    The base class resolves the bind address with socket.getfqdn() between
+    bind() and listen(). On macOS that lookup of 127.0.0.1 has no answer and
+    times out after ~35s, delaying every sidecar start by that much before the
+    facade can reach it. server_name only fills in CGI variables, unused here.
+    """
+
+    def server_bind(self):
+        socketserver.TCPServer.server_bind(self)
+        self.server_name, self.server_port = self.server_address[:2]
+
+
 def main() -> int:
     if PORT == 0:
         print("self-audit: $SIDECAR_PORT not set", file=sys.stderr)
         return 2
-    srv = ThreadingHTTPServer(("127.0.0.1", PORT), Handler)
+    srv = Server(("127.0.0.1", PORT), Handler)
     print(
         f"[self-audit] listening on 127.0.0.1:{PORT} skill={SKILL} "
         f"secret={fingerprint(SECRET)}",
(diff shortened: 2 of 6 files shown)


nhuelstng:
## macOS verification (CI run [30946126718](https://github.com/TNG/oh-my-agentic-coder/actions/runs/30946126718))

`E2E: model-free (macos-latest)` — the step that carried the stall:

| step | before | after |
|---|---|---|
| Check model probe against a stub gateway | **659s** | **27s** |
| whole job | 700s | 75s |

27s is the floor: 20s of it is case 10's deliberate retry sleeps, the rest is
the 69 requests and 17 restarts at their real cost (~110ms per restart now
that the lookup is gone).

`Test (macos-latest)` went 132s → 116s, its `go test` step 87s → 71s, which is
the facade integration test no longer paying one 35s lookup (it spawns the
sidecar less often than the probe test restarts its stub).

All 10 checks green, both OSes.

Issue #210, opened 2026-08-07
harden(security): authenticate the omac control plane (deactivate DoS; per-caller identity)


nhuelstng:
### Duplicate check
- [x] I have searched the existing issues

## Context / Summary
Follow-up to #208. The `serve`/`start` control plane (`OMAC_CONTROL_BASE`) is injected into the sandbox and is unauthenticated.

## Problem / What
- `serve`'s `/__omac__/deactivate` calls `StopSidecar` and is reachable from inside the sandbox, so a confined agent can stop its own sidecars (self-DoS) and — if it knows another directory's path — tear down another directory's sidecars. The plugin legitimately needs activate/deactivate, so it cannot simply be blocked.
- More broadly there is no per-caller identity on the control plane, so one sandbox can act on another's directory (see also the prior dir_token isolation work).
- Minor: approvals are additive and never auto-retired, allowing rollback to a previously-approved build; and the approval store must not sit under a sandbox-granted tree (`XDG_CONFIG_HOME`).

## Suggested fix / Ask
- [ ] Authenticate spawn/teardown-causing control endpoints with a token available only outside the sandbox, or bind each request to the calling sandbox's identity.
- [ ] Consider retiring superseded approvals on `register --force` to remove the rollback surface.
- [ ] Assert the approval store path resolves outside every granted tree; warn otherwise.

## Non-goals
- Removing the control plane (the plugin needs it).

Issue #208, opened 2026-08-07
fix(security): sandboxed agent can self-author + spawn a skill sidecar outside the sandbox


nhuelstng:
### Duplicate check
- [x] I have searched the existing issues

## Context / Summary
Skill sidecars run **unsandboxed** (the supervisor spawns them as plain host children — deliberate, since the marketplace needs host network + secrets). But the workdir is agent-writable, including skill source dirs (`.opencode/.claude/.agents/skills/*`) and the workdir registry (`.opencode/sidecar.json`). Nothing anchored the decision to spawn a skill outside that agent-writable area.

## Problem / What
A confined agent can author a skill, get it registered (or forge `.opencode/sidecar.json`), trigger a spawn (`omac register` from inside the sandbox, or the live-reload/serve auto-register path), and run its own code **outside the sandbox** — reading host files and reaching the network the sandbox denies it. Editing an already-trusted skill's on-disk sidecar code is the same escape by another route.

Empirically reproduced (pre-fix): an agent-authored skill's sidecar returned a file from outside the workdir through the facade the agent reaches.

Related process-manipulation surface (verified, mostly already closed by the sandbox): the agent **cannot** signal/ptrace/read-memory of a sidecar (`--unshare-pid` + fresh `/proc`, unprivileged user-ns, no caps) nor reach its loopback port (Landlock TCP rules). The one residual is the `serve` control plane's unauthenticated `/__omac__/deactivate`, which can stop a sidecar.

## Suggested fix / Ask
Anchor the spawn decision in a host-only store the sandbox cannot write, keyed by content hash:
- [x] `internal/skilltrust`: approvals under `~/.config/omac/approvals.json`, keyed by (skill name, bundle hash)
- [x] Gate every spawn path (start cold-start, start live-reload, serve activation) + supervisor authorizer backstop
- [x] `omac register` approves (host-side only); `omac deregister` revokes the exact (name, hash)
- [x] Trust-on-first-upgrade grandfathering, one-shot
- [x] Docs in `docs/SECURITY_MODEL.md`; unit + e2e tests

## Non-goals
- Sandboxing the sidecars themselves (breaks the marketplace's host access).
- Closing the residuals below — tracked separately.

## Evidence / Environment
`go test ./internal/...` and `go test -tags=e2e_fast ./internal/e2e/` green; `TestE2ESelfAuthoredSkillRefused` fails (`ESCAPE: sidecar RAN`) when the gate is disabled.

PR #212, opened 2026-08-07
harden(security): execute skill sidecars from an approved snapshot

### LINKED ISSUE #209: harden(security): execute skill sidecars from an approved immutable snapshot
### Duplicate check

## Context / Summary
Follow-up to #208. The host-only approval gate binds a skill to its `config.BundleHash`. That hash is only as tight as what it covers.

## Problem / What
`config.BundleHash` (`internal/config/meta.go`) deliberately excludes dependency/artifact subtrees (`node_modules`, `.venv`, `dist`, `build`, `target`, …) and skips symlinks, so an install can populate them without churning the hash. Consequence: an **approved** skill whose executed code lives under an excluded dir (`node dist/index.js`, a vendored `.venv`) or behind a symlink into agent-writable space can be swapped **after approval** without invalidating it — unsandboxed code execution. This bound applies equally to the existing bundle-hash drift check.

## Suggested fix / Ask
- [ ] At approval time, snapshot the skill dir to an immutable host-side location (under `~/.config/omac`, not mounted into the sandbox) keyed by bundle hash.
- [ ] Spawn the sidecar from that snapshot (`cmd.Dir` = snapshot), so executed bytes == approved bytes regardless of later workdir edits. Also closes the TOCTOU between hash and exec.
- [ ] Handle skills that write to their own dir at runtime; GC snapshots on deregister.

## Non-goals
- Changing `BundleHash`'s exclusion set repo-wide (would break install-churn stability of the drift check).

nhuelstng:
**Issue:** Closes #209

> **Stacked on #211** (base branch `fix/skill-spawn-approval-anchor`). Review/merge #211 first; this targets that branch, so the diff here is only the snapshot layer.

## What
- Freeze a skill's whole tree into a host-only, content-addressed snapshot at approval time (`~/.config/omac/skills/<name>/<hash>`), with in-tree symlinks flattened to their target content.
- Spawn the sidecar **from the snapshot**, not the agent-writable workdir (all paths: start cold-start, live-reload, serve).
- `skilltrust.Approve` creates the snapshot; `Revoke` removes it.

## Why
The approval was keyed on `config.BundleHash`, which excludes `node_modules`/`.venv`/`dist`/`build`/`target` and skips symlinks. So an approved skill executing code from an excluded tree (`node dist/index.js`, a vendored `.venv`) or via a symlink could be tampered with in the workdir **after** approval without changing its hash — plus a check-then-exec TOCTOU. This was the top residual from #208's review.

## How
The snapshot is the enforcement boundary: executed bytes == approved bytes, captured when a host-side actor approves. `~/.config/omac` is never mounted into the sandbox, so the agent cannot touch the snapshot. `approvedSpawnDir(name, workdirDir)` hashes the workdir content, matches the approval, and returns the frozen snapshot dir to spawn from. Snapshots are content-addressed (identical content stored once; re-approval idempotent) and GC'd on deregister.

## Verification
- `go build ./...`, `go vet ./internal/...` — clean.
- `go test $(go list ./internal/... | grep -v /internal/e2e)` — green.
- `go test -tags=e2e_fast -count=1 ./internal/e2e/` — green.
- **`TestSnapshotDefeatsPostApprovalTampering`**: approve a skill reading a `.venv/` payload (excluded from the hash), rewrite that payload in the workdir (hash unchanged), reload — the sidecar serves the **frozen** payload, not the tamper.
- `TestSnapshotFreezesContentAndRevokeRemovesIt`: snapshot is immutable to source edits and removed on revoke.

## Follow-up
- #210 — authenticate the control plane (serve `deactivate` DoS; per-caller identity; approval rollback/XDG notes) — still open.

### DIFF
--- a/CREATING_A_SKILL.md
+++ b/CREATING_A_SKILL.md
@@ -183,6 +183,18 @@ Two paths to be aware of:
   not the bare filename. Compiled-language skills typically have the
   install script drop a binary at `scripts/sidecar` and reference
   `["./scripts/sidecar"]`.
+
+  > **The cwd is an immutable snapshot, not your workdir.** After a skill is
+  > approved (`omac register` on the host), omac freezes its tree into a
+  > host-only snapshot and spawns the sidecar from *that* copy — so the cwd is
+  > read-only-in-spirit and is NOT the project directory. Do not write files
+  > relative to cwd expecting the agent (or the project) to see them: those
+  > writes land in the snapshot, are invisible to the sandboxed agent, and do
+  > not persist across a re-approval. Exchange data through your HTTP endpoints
+  > (the facade) and write to `OMAC_WORKDIR` when you need the project dir.
+  > For the same reason, vendor/install dependencies BEFORE `omac register`
+  > (or install them system-wide) — deps added after approval are not in the
+  > snapshot. See [`docs/SECURITY_MODEL.md`](./docs/SECURITY_MODEL.md#self-authored-skills).
 - **`scripts/` is shared between two consumers**: the *omac supervisor*
   spawns the sidecar entry-point as a long-running HTTP service outside
   the sandbox; the *agent* may also invoke other helper scripts in this
--- a/docs/SECURITY_MODEL.md
+++ b/docs/SECURITY_MODEL.md
@@ -123,16 +123,16 @@ the sandbox:
 
 - `config.BundleHash` deliberately excludes dependency/artifact subtrees
   (`node_modules`, `.venv`, `build`, `dist`, `target`, …) and does not follow
-  symlinks — so an install can populate those after registration without
-  invalidating the approval. A skill that executes mutable code from an
-  excluded tree (e.g. `node dist/index.js`, a vendored `.venv`) or via a
-  symlink into agent-writable space can therefore be tampered with after
-  approval without changing its hash. This bound applies equally to the
-  existing bundle-hash drift check. The complete fix is to execute the
-  sidecar from an immutable, host-side snapshot taken at approval time
-  (so the executed bytes are exactly the approved bytes); that is tracked as
-  follow-up hardening. Until then, the approval binds only the hashed subset
-  of a skill's tree.
+  symlinks, so the hash alone does not cover everything a skill executes.
+  omac closes this by **executing from an approval snapshot**: at approval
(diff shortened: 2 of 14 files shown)


nhuelstng:
## Adversarial review pass (4 reviewers: security, breakage/regression, complexity, UX)

Addressed in `f36c736`:
- **Security (real regression, now fixed):** `copyTree` dereferenced symlinks and baked the target's *content* into the snapshot — a planted `dist/x -> ~/.ssh/id_rsa` (excluded from the bundle hash, so hash unchanged) copied a host secret into `~/.config/omac` and let a file-serving skill read it back. Now symlinks are never dereferenced (in-tree relative links recreated; escaping/absolute dropped). New `TestSnapshotDoesNotBakeEscapingSymlink`.
- **Coupled correctness bug:** the supervisor backstop re-hashed the snapshot (which now differs from the workdir hash for symlinked skills) → false refusal. Replaced with `skilltrust.IsSnapshotPath` (spawn only from the host-only snapshot root — unforgeable, no re-hash).
- **UX ("looks broken"):** `omac register` no longer prints green `[ok]` when approval fails (in-sandbox case) — prints a `[warn]` with the host-terminal remedy; `omac serve` now announces first-upgrade grandfathering; start live-reload reports an unapproved skill to the human (was `--verbose`-only), deduped; `CREATING_A_SKILL.md` corrected (cwd is the immutable snapshot, not the workdir).

Deferred to follow-ups:
- #215 — snapshot lifecycle: GC superseded snapshots on re-register; deps-installed-after-register.
- #210 — control-plane auth; also covers approval rollback (additive, never retired) and the grandfather-workdir trust-on-first-use note.

Non-blocking nice-to-haves (not done, low value): unexport `skilltrust` symbols with no external callers (`Load`/`Store`/`Approval`/`Path`/`Snapshot`/`SchemaVersion`); trim duplicated threat-model prose to the one canonical `SECURITY_MODEL.md`; add distinct `X-Omac-Reason` codes (`skill-unapproved`/`skill-drift`/`skill-health`) so consumers branch on a code rather than `detail` text.

Verdict from the panel: the combined design is sound and minimal (snapshot is the right mechanism vs. hashing-more/read-only-bind); the security regression above was the one must-fix and is resolved.

Issue #176, opened 2026-07-25
facade.New's interface is 7 positional params plus 10 mutable fields, and the only way in is a real listener


nhuelstng:
## Duplicate check

- [x] I have searched the existing issues

## Context / Summary

Found during an architecture review of `main` (`6257211`).

Everything a caller must know about the facade lives *outside* its constructor, and the only way into the request path is a real listener. The tests show the cost: 17 sites build `&Facade{…}` by hand, in states `New` can never produce, and three reach the unexported handler.

This is a prerequisite for #88 (drop the TCP transport): "at least one listener configured" is currently validated at `Start`, per-listener, with `""` meaning "disabled" — a `Config` with validation at construction makes that change local instead of scattered.

### The interface is ~40 facts wide

`internal/facade/facade.go:176`:

```go
func New(socketPath, tcpAddr string, routes []Route, maxBody int64, idle time.Duration, accessLog, version string) *Facade
```

7 positional params — two adjacent `string`s that are easy to transpose, and five bare primitives where `""` / `0` silently means "disabled" or "inherit". On top of that, `internal/facade/facade.go:124-192` exposes **10 mutable exported fields**, three of which change endpoint behaviour when nil:

| field | nil / zero means |
|---|---|
| `ProtectedPathChecker` | `GET /sandbox/denied` 404s (see #173) |
| `IntentRegistry` | `POST /sandbox/intent` 503s |
| `DenialNote` | falls back to `sandboxdeny.Default().FacadeNote` |
| `MaxBodyBytes` / `IdleTimeout` on `Route` | inherit from the facade |
| `SkillDir` on `Route` | SKILL.md discovery disabled |
| `State` on `Route` | treated as `RouteReady` |

Plus ordering and ownership rules the signature cannot express:

- `SetAuditor` (`:200`) is "safe to call before `Start`"; `f.auditor` is then read unsynchronised from request goroutines at `:901`.
- `Close()` closes `IntentRegistry` (`:355`) — a registry the facade did not create (it is built in `internal/cli/facade_wiring.go:34`).
- `AddRoute` (`:210`) does not update the exported `Routes` slice; `Routes` is written once in `New` (`:185`) and **read nowhere in the tree** (verified: `grep -rn "\.Routes"` returns no readers). It desyncs from the live `routes` map on the first `AddRoute`.
[...]

Issue #175, opened 2026-07-25
runLaunch holds the launch lifecycle in locals; the strict-audit abort path skips four cleanups including secret zeroing


nhuelstng:
## Duplicate check

- [x] I have searched the existing issues

## Context / Summary

Found during an architecture review of `main` (`6257211`). `internal/cli/start.go` is the highest-churn file in the repo (20 of the last 120 commits).

`runLaunch` is a **769-line single function** that holds an entire process lifecycle in local variables. Because no object owns the acquired resources, teardown is written twice — and the two versions disagree. The code's own comment asserts they agree.

Refs #174 (the 123-line arming phase inside `runLaunch`), #173 (the facade-wiring phase). Both shrink this function; doing them first makes this extraction much smaller.

## Problem / What

`internal/cli/runLaunch` — `internal/cli/start.go:180-948`:

- 769 lines, 17 unnamed phases (comment markers at `:205, 224, 231, 247, 278, 316, 351, 592, 603, 673, 707, 729, 758, 775, 848, 896, 910`)
- 26 `return Exit*` sites, 8 `defer`s (`:376, 637, 649, 668, 675, 700, 751, 770`), 18 rebinds of one shared `err`
- ~65 locals still live at the `sandbox.ExecWithReady` call (`:936`)
- begins by re-unpacking `opts` into 9 shadow locals (`:181-189`), so both `opts.noSandbox` and `noSandbox` stay live for 750 lines
- `internal/cli/runServe` (`serve.go:48-530`, 483 lines) re-derives much of the same sequence

### The concrete defect: the strict-audit abort skips four cleanups

`fatalTeardown` (`start.go:900-908`) ends in `os.Exit`, which skips every `defer`:

```go
fatalTeardown = func(ferr error) {
	fmt.Fprintln(env.Stderr, prefix+": audit (strict) write failed, aborting:", ferr)
	if !keepRunning { sup.ShutdownAll(5 * time.Second) }
	_ = f.Close()
	closeControl()
	os.Exit(ExitIOError)
}
```

Diffed against the deferred stack, it omits:

| skipped cleanup | defined at |
|---|---|
| `allSecrets[i].Zero()` — **plaintext secrets are not wiped** | `start.go:376-380` |
| `os.RemoveAll(sandboxTmp)` — temp dir leaked | `:649` |
| `cacheScope.Close()` — releases a flock (`toolcache/cache.go:131`) | `:668` |
[...]

Issue #174, opened 2026-07-25
Skill readiness (secrets + config resolution) is implemented five times and the copies have diverged


nhuelstng:
## Duplicate check

- [x] I have searched the existing issues

## Context / Summary

Found during an architecture review of `main` (`6257211`). Sits on the hottest churn path in the repo: `internal/cli/start.go` (20 commits) and `internal/cli/serve.go` (13) over the last 120 commits.

"Is this skill ready to mount?" — resolve its declared secrets and config fields, then decide — is answered in **five** places. The copies have already diverged, and the divergences are silent.

Related but distinct: #136 (auto-*registering* skills on start) changes the gate *before* this resolution; this issue is about the resolution itself.

## Problem / What

Five implementations of the same rule:

| # | Copy | Lines |
|---|---|---|
| 1 | `internal/cli/start.go:416-538` (+ refusal renderer `:542-590`) | 123 |
| 2 | `internal/cli/serve.go:1277-1340` | 64 |
| 3 | `internal/cli/start_reload.go:228-272` | 45 |
| 4 | `internal/cli/doctor.go:77-130` | 54 |
| 5 | `internal/cli/config_cmd.go:235-272` (config only) | 38 |

Verified behavioural divergences:

| copy | `$spec.DefaultFromEnv` | `env_passthrough` secret fallback | distinguishes keychain I/O error from not-found | on missing |
|---|---|---|---|---|
| `start.go:504` / `:469` | yes | yes (`:480`) | yes (`:471`) | consolidated refusal + exit code |
| `serve.go:1328` / `:1300` | yes | **no** | yes (`:1305`) | route `pending-credentials` |
| `start_reload.go:257` / `:242` | **no** | **no** | **no** | **silent `continue`** |
| `doctor.go:114` | no | no | n/a | **unscoped** `keychain.Has` |

### Failure 1 — live reload silently never mounts a skill

`internal/cli/start_reload.go:257-266` omits `spec.DefaultFromEnv` entirely, and `:242-253` treats *any* keychain error as "secret absent". A skill declaring

```yaml
config:
  - name: API_BASE
    default_from_env: MY_API_BASE
    required: true
```

starts fine under `omac start` (`start.go:519`) and `omac serve` (`serve.go:1328`), but the live-reload loop `continue`s past it forever — no message, no route, no diagnostic.

### Failure 2 — `serve` rejects credentials the sidecar would have received

[...]

PR #219, opened 2026-08-11
fix(cli,sandbox): resolve the sandbox plan once so /sandbox/denied answers

### LINKED ISSUE #173: Facade sandbox wiring resolves the launcher profile name against the sandbox-policy namespace, disabling GET /sandbox/denied on every default launch
## Duplicate check

## Context / Summary

Found during an architecture review of `main` (`6257211`), scoped to the highest-churn packages. This is the only *bug* of the four findings — the rest are structural. Filed as the smallest independent fix.

`omac` has **two unrelated namespaces both called "sandbox profile"**:

| namespace | keyed by | example keys | resolver |
|---|---|---|---|
| launcher profile — a templated argv | `config.LauncherConfig.Sandbox.Profiles` | `builtin`, `nono`, `no-sandbox-debug` | map index, `internal/cli/start.go:218` |
| sandbox policy — the grant JSON | `~/.config/omac/sandbox-profiles/<ref>.json` | `default` | `sandboxprofile.Resolve`, `internal/sandboxprofile/resolve.go:146` |

Both are plain `string`, so the compiler cannot tell them apart. One call site passes the wrong one.

## Problem / What

`internal/cli/facade_wiring.go:23-35` takes a bare `profName` and feeds it to the *policy* resolver:

```go
func wireFacadeSandbox(f *facade.Facade, noSandbox bool, profName string, warn func(format string, args ...any)) {
	if !noSandbox {
		if prof, _, err := sandboxprofile.Resolve(profName); err == nil {   // profName == "builtin"
			f.ProtectedPathChecker = sandboxrun.NewProtectedPathSet(prof)
			...
```

But both callers pass the **launcher** name:

- `internal/cli/start.go:744` — `profName` from `lc.Sandbox.DefaultProfile` / `--sandbox` (`start.go:214-221`)
- `internal/cli/serve.go:258` — same

The default launcher profile is `builtin` (`internal/config/launcher.go:137,170`), whose *policy* ref is `default`, spelled inside its command template (`launcher.go:172-174`: `"--profile", "default"`). `Resolve("builtin")` therefore looks for a file that does not exist and errors, because `resolve.go:162-163` only scaffolds for `ref == "default"`.

Consequences on **every default launch** (`omac start`, `omac serve`, no flags):

- `f.ProtectedPathChecker` stays nil → `GET /sandbox/denied` returns 404 (`internal/facade/facade.go:521`). The agent cannot tell a sandbox denial from a genuinely missing file — the endpoint's whole purpose.
- `denial.facade_note` from the profile is never applied (`facade_wiring.go:27-29`).
- The user sees the `warn(...)` line on stderr on every launch.

The correct launcher→policy mapping already exists and is used elsewhere: `inspectBuiltinProfileRef(prof.Command)` (`internal/cli/doctor.go:299-318`), consumed by `sandboxProfileAllowVarsEmpty` (`internal/cli/serve.go:682`) with the read-only resolver. `wireFacadeSandbox` simply never receives `prof`.

Secondary, same seam: `Resolve` **writes** `default.json` as a side effect of reading (`resolve.go:168`), while its near-identical twin `ResolveReadOnly` (`resolve.go:109-135`) does not. So wiring an HTTP endpoint and running `omac provenance` both scaffold user files. The two functions differ in exactly one branch.

There is **no** `internal/cli/facade_wiring_test.go`, and nothing in `internal/cli` or `internal/e2e` asserts a non-nil `ProtectedPathChecker` after a launch — only `internal/facade/sandbox_denied_test.go`, which injects a stub.

## Suggested fix / Ask

Make the seam a type so the wrong call stops compiling.

- [ ] Introduce a resolved value carrying both namespaces, e.g. `SandboxPlan{LauncherName string; Launcher config.SandboxProfile; Native *sandboxprofile.Profile; NativePath string}`, produced once per launch by a single `ResolveSandboxPlan(lc, flagProfile, noSandbox)`.
- [ ] Move `inspectBuiltinProfileRef` out of `cli/doctor.go` into the plan resolver (or onto `config.SandboxProfile` as `PolicyRef() (ref string, native bool)`).
- [ ] Change `wireFacadeSandbox` to take the plan (or the already-resolved `*sandboxprofile.Profile`) instead of a name.
- [ ] Have `sandboxProfileAllowVarsEmpty`, `profileRunsNativeSandbox` and `forwardHarnessEnv` read the plan instead of re-resolving from a string.
- [ ] Collapse `Resolve` / `ResolveReadOnly` into `Resolve(ref string, opts ...ResolveOption)` with `WithScaffold()`; **read-only by default**. Only first-run setup opts in.
- [ ] Add `facade_wiring_test.go` asserting a non-nil `ProtectedPathChecker` for the default profile, plus an e2e assertion that `GET /sandbox/denied?path=…` answers under a default launch.

### Acceptance criteria

- [ ] A default `omac start` / `omac serve` wires a non-nil `ProtectedPathChecker`; `GET /sandbox/denied` answers instead of 404.
- [ ] `denial.facade_note` from the resolved profile reaches `f.DenialNote`.
- [ ] Passing a launcher profile name where a policy ref is expected is a compile error, not a runtime warning.
- [ ] Resolving a profile for inspection creates no files.
- [ ] Regression test covers the default-profile path (currently untested).

## Non-goals

- Changing the grant model, the profile JSON schema, or the `omac sandbox run` flag vocabulary.
- Renaming the user-facing `--sandbox` flag or the launcher profile keys.
- Reworking how grants are spliced into the sandbox argv (separate concern).

## Evidence / Environment

`main` @ `6257211`, Linux. Reproduced twice — through the raw resolver and through the real `DefaultLauncherConfig`:

```go
// internal/sandboxprofile — probe
t.Setenv("XDG_CONFIG_HOME", t.TempDir())
Resolve("builtin")   // -> ERROR: sandbox profile "builtin" not found
                     //    (expected ~/.config/omac/sandbox-profiles/builtin.json)
Resolve("default")   // -> ok
```

```go
// internal/cli — probe, real default config
lc := config.DefaultLauncherConfig()
profName := lc.Sandbox.DefaultProfile              // "builtin"
f := facade.New("", "127.0.0.1:0", nil, 0, 0, "", "probe")
wireFacadeSandbox(f, false, profName, collectWarning)
// warnings=1  ProtectedPathChecker==nil? true  DenialNote=""
```

Both probes were run against the checked-out tree and then deleted; no test files were left behind.

🤖 Filed by an agent during an architecture review of `main`.

Retr0MrWave:
**Issue:** Closes #173

## What

- The launcher/policy namespace split becomes one resolved value per launch (`sandboxPlan`), so a launcher profile name can no longer be handed to the policy resolver — it is a compile error.
- `GET /sandbox/denied` answers on a default `omac start` / `omac serve` instead of 404, and `denial.facade_note` reaches the facade.
- `sandboxprofile.Resolve` is read-only by default (`WithScaffold()` on the launch path only), so inspecting a profile no longer writes `default.json`.
- First regression coverage for the default-profile path: `facade_wiring_test.go`, `sandboxplan_test.go`, `PolicyRef` table tests, and a model-free `e2e_fast` test that drives the real binary.

## Why

`wireFacadeSandbox` took a bare profile name and fed it to the **policy** resolver, while both callers passed the **launcher** name. The default launcher profile is `builtin`; its policy ref is `default`, spelled inside the command template (`internal/config/launcher.go:174`). `Resolve("builtin")` therefore looked for a file that never exists, on **every default launch**:

- `ProtectedPathChecker` stayed nil → `GET /sandbox/denied` returned 404, so the agent could not tell a sandbox denial from a genuinely missing file — the endpoint's entire purpose;
- `denial.facade_note` was never applied;
- a warning was printed on stderr every time.

Both namespaces are plain `string`, so the compiler could not tell them apart. Nothing in `internal/cli` or `internal/e2e` asserted a non-nil checker after a launch.

## How

- `config.SandboxProfile.PolicyRef() (ref string, native bool)` owns the launcher→policy mapping — moved from the untested `cli.inspectBuiltinProfileRef`, and it subsumes `profileRunsNativeSandbox` (identical anchor check), which is deleted.
- `internal/cli/sandboxplan.go`: `resolveSandboxPlan(lc, flagProfile)` resolves the launcher profile and, when native, its policy — once per launch, read-only. An unknown launcher name is returned as an error; callers keep today's fatality rule (not fatal under `--no-sandbox` / `--no-inner`). A missing policy is recorded in `PolicyErr`, never fatal — the `omac sandbox run` child resolves it itself.
- `wireFacadeSandbox`, `forwardHarnessEnv` and `injectSandboxEnvAllow` take the plan; `serve.go`'s two duplicated resolver helpers collapse into `planAllowVarsEmpty` / `planDeniedBaseVars`. The two env warnings now name the policy ref (the file the user edits) rather than the launcher name.
- `Resolve(ref string, opts ...ResolveOption)` replaces the `Resolve` / `ResolveReadOnly` pair, which differed in exactly one branch. Only `internal/sandboxrun/run.go` passes `WithScaffold()`, so first launch still creates the user's editable `default.json` — from the child rather than from the parent's HTTP wiring.

Second commit addresses review findings — two latent inconsistencies that a live `ProtectedPathChecker` made reachable, plus the last instance of the conflation:

- `omac serve --learn` lifts every filesystem restriction in the child (`Grants.withUnrestrictedFilesystem` drops `ProtectedPaths` and grants `/`), but the facade would still report the profile's static set, telling the agent a genuinely missing file was blocked. The wiring now takes the learn flag and installs `sandboxrun.UnrestrictedProtectedPathSet()`.
- Read-only `Resolve` can return an empty profile path (compiled-in defaults, no file consulted); `PagesPath("")` turned that into the **relative** `.pages.json`, so `omac provenance` in a directory holding such a file would report it as the profile's learned network decisions. Fixed at the root, which also covers the pre-existing caller in `diagnose_probe.go:41`.
- `doctorProfileLint` hard-coded the `"default"` policy ref; it now follows the default launcher template via `defaultPolicyRef`.

## Verification

```
go build ./... && go vet ./... && go vet -tags=e2e_fast ./internal/e2e/     # clean
go test -race -count=1 ./internal/...                                      # pass *
go test -tags=e2e_fast -race -count=3 -timeout=8m ./internal/e2e/          # ok (4.4s)
```

The third commit fixes a defect in the second: the new e2e test read its
stderr `bytes.Buffer` while `os/exec`'s copier goroutine was writing to it, so
the model-free CI leg — which runs with `-race` — reported a data race instead
of the assertions. It passed locally only because I had not used `-race`. Now
reproduced and verified with CI's exact invocation, repeated three times.

\* `internal/sandboxrun` has two failures on my host that are identical on `main` (`TestIntegrationWorktreeKnownLimitations` — git `packed-refs`/`Device or resource busy`; `TestIntegrationWorkflowInterpretersRunnable` — missing interpreters). Verified by stashing the change and re-running.

Every new guard was checked **red** against the unfixed code — the pre-fix wiring, `learnMode` ignored, and the unguarded `PagesPath`:

```
--- FAIL: TestWireFacadeSandboxDefaultProfileWiresChecker
    ProtectedPathChecker is nil; GET /sandbox/denied would 404 on a default launch
--- FAIL: TestWireFacadeSandboxAppliesDenialFacadeNote
--- FAIL: TestE2ESandboxDeniedAnswersOnDefaultLaunch
--- FAIL: TestWireFacadeSandboxLearnModeProtectsNothing
--- FAIL: TestPagesPathEmptyProfilePath
```

Manual end-to-end against the compiled binary, fresh `HOME`, default launch (no flags):

```
$ curl -si ".../sandbox/denied?path=$HOME/.aws/credentials"
HTTP/1.1 200 OK
{"denied":true,"path":"…/.aws/credentials","rule":"baseline","note":"Intentionally restricted…"}   # was 404
$ curl -si ".../sandbox/denied?path=$HOME/main.go"    -> 404 {"denied":false,…}
$ grep -c "could not be resolved" serve.stderr        -> 0
$ HOME=$fresh omac provenance && ls $fresh/.config    -> no such directory   # inspection writes nothing
$ HOME=$fresh omac sandbox run -- /bin/true && ls …/sandbox-profiles/  -> default.json   # launch still scaffolds
```

Linux only; the macOS Seatbelt path is unaffected by this change (the new e2e test runs with `--no-inner`, so it is sandbox-free) and is covered by the `e2e.yml` matrix.

CI on this branch: all legs green (`E2E: model-free` ubuntu + macOS, `Test`
ubuntu + macOS, `Cache integration`, `External tooling`, `Lint`, `WSL2-style
session`, DCO).

## Follow-up

- Named types for the refs themselves (`config.LauncherProfileName`, `sandboxprofile.Ref`) would make the two namespaces unmixable everywhere, not just at this seam. Left out deliberately: it ripples into the launcher YAML map key, `sandboxprofile.Flags`/`ParseFlags` and the audit event fields.
- #174, #175, #176 are the remaining findings from the same architecture review.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -109,7 +109,9 @@ environment variables.
 The built-in sandbox reads JSON profiles from
 `~/.config/omac/sandbox-profiles/`. On first `omac start` with the
 `builtin` profile, omac scaffolds `default.json` from the compiled-in
-defaults so you can edit it:
+defaults so you can edit it (only the launch itself writes it — the
+inspection commands `omac doctor`, `omac diagnose` and `omac provenance`
+read the compiled-in defaults instead of creating files):
 
 ```
 ~/.config/omac/sandbox-profiles/
--- a/internal/cli/diagnose.go
+++ b/internal/cli/diagnose.go
@@ -45,7 +45,7 @@ func runDiagnose(args []string, env *Env) int {
 		return ExitMisuse
 	}
 
-	profile, profPath, err := sandboxprofile.ResolveReadOnly(*profileRef)
+	profile, profPath, err := sandboxprofile.Resolve(*profileRef)
 	if err != nil {
 		fmt.Fprintln(env.Stderr, "omac diagnose:", err)
 		return ExitConfigInvalid
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -160,8 +160,10 @@ func runDoctor(args []string, env *Env) int {
 
 	// Static security lint of the resolved profile (advisory). Reuses the
 	// same engine as `omac provenance --check` — findings are warnings
-	// here, never a doctor failure.
-	doctorProfileLint(env, "")
+	// here, never a doctor failure. The ref follows the default launcher
+	// profile's template, so a config pointing at a non-default policy gets
+	// that policy linted rather than an unused "default".
+	doctorProfileLint(env, defaultPolicyRef(lc))
 
 	fmt.Fprintln(env.Stdout, "\nWhen a run fails, `omac diagnose` shows what the sandbox blocked and why.")
 
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -176,7 +176,7 @@ func runDoctor(args []string, env *Env) int {
 // the gap where doctor never surfaced the security lint (previously only
 // reachable via `omac provenance --check`).
 func doctorProfileLint(env *Env, profileRef string) {
-	profile, _, err := sandboxprofile.ResolveReadOnly(profileRef)
+	profile, _, err := sandboxprofile.Resolve(profileRef)
 	if err != nil {
 		return // profile problems are already reported by the sandbox section
 	}
@@ -282,13 +282,13 @@ func doctorSandboxProfileWarnings(env *Env, lc config.LauncherConfig) {
 		if len(prof.Command) == 0 {
 			continue
 		}
-		ref, ok := inspectBuiltinProfileRef(prof.Command)
-		if !ok {
+		ref, native := prof.PolicyRef()
+		if !native {
 			// Opaque external launcher (nono, no-sandbox-debug, etc.):
 			// doctor can't see into its profile, so skip silently.
 			continue
 		}
-		p, _, err := sandboxprofile.ResolveReadOnly(ref)
+		p, _, err := sandboxprofile.Resolve(ref)
 		if err != nil {
 			fmt.Fprintf(env.Stdout, "  [warn] sandbox profile %q: %v\n", profName, err)
 			continue
@@ -322,36 +322,6 @@ func doctorSandboxProfileWarnings(env *Env, lc config.LauncherConfig) {
 			fmt.Fprintf(env.Stdout, "         remediation: %s\n", w.remediation)
 		}
 	}
-}
-
-// inspectBuiltinProfileRef looks at a sandbox profile Command argv
-// template and, if it is a {{self}} sandbox run invocation, extracts
-// the --profile reference. Recognized run forms:
-//   - "--profile", "default"   (separate args)
-//   - "--profile=default"      (inline)
-//   - omitted --profile        (resolves to "default")
-//
-// Only {{self}} sandbox run commands are inspectable; other sandbox
-// subcommands and external launchers are opaque and return ok=false.
-func inspectBuiltinProfileRef(command []string) (string, bool) {
-	if len(command) < 3 || command[0] != "{{self}}" || command[1] != "sandbox" || command[2] != "run" {
-		return "", false
-	}
-	// Find "--profile" (separate or inline) before "--".
-	for i := 3; i < len(command); i++ {
-		arg := command[i]
-		if arg == "--" {
-			break
-		}
-		if arg == "--profile" && i+1 < len(command) {
-			return command[i+1], true
-		}
-		if strings.HasPrefix(arg, "--profile=") {
-			return strings.TrimPrefix(arg, "--profile="), true
-		}
-	}
-	// Omitted --profile resolves to "default".
-	return "default", true
 }
 
 // profileGrantWarnings returns warnings for broad tool-home and
--- a/internal/cli/facade_wiring.go
+++ b/internal/cli/facade_wiring.go
@@ -21,9 +21,17 @@ import (
 // disabling the endpoint. The intent registry is always wired: in-memory,
 // session-scoped, written by the agent via POST /sandbox/intent and read
 // by the popup via GET.
-func wireFacadeSandbox(f *facade.Facade, noSandbox bool, plan sandboxPlan, warn func(format string, args ...any)) {
+//
+// learnMode (`omac serve --learn`) is a launch fact the policy file cannot
+// carry: it lifts every filesystem restriction in the child, so the static
+// protected set must not be reported for that session.
+func wireFacadeSandbox(f *facade.Facade, noSandbox, learnMode bool, plan sandboxPlan, warn func(format string, args ...any)) {
 	if !noSandbox {
 		switch {
+		case learnMode:
+			// Nothing is protected in a learn session; say so rather than
+			// claiming the profile's static set is in force.
+			f.ProtectedPathChecker = sandboxrun.UnrestrictedProtectedPathSet()
 		case plan.Policy != nil:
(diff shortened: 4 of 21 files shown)


JonasKellerer (commented):
Thank you for the PR. This PR looks good to me, but I am new to the project. So I am not sure if I should/can approve it. @nhuelstng 

I've setup omac and verified that at startup there is no longer the warning that "builtin" can longer be resolved."

Also the tests are running.

nhuelstng:
@JonasKellerer its totally fine if you approve it :)

JonasKellerer (approved):

PR #228, opened 2026-08-13
fix(cli): suggest re-register commands that survive a follow-up flag

### LINKED ISSUE #227: Small display bug: instruction to re-register skills are wrong
### Duplicate check

### Context / Summary

omac suggests a wrong shell command

### Problem / What

omac suggests to re-register a skill as shown in the screenshot:
`omac register --force skill-marketplace` (or maybe even `skill-marketplace - omac register --force skill-marketplace`)
It should be `omac register skill-marketplace --force` though

Image

### Suggested fix / Ask

Suggested fix would be to show the correct shell command.

### Non-goals

_No response_

### Evidence / Environment

_No response_

willm-TNG:
**Issue:** Closes #227

## What

- Argument reordering is now boolean-aware, so a bool flag written before a positional no longer causes every later flag to be dropped.
- The four skill-problem hints in `start`/`serve` render through a shared helper, labelled and in the documented `omac register <skill> [flags]` order.
- Docs and the bundled `omac-write-a-skill` copy of the skill guide updated to match.

## Why

#227 reported that `omac register --force <skill>` looks wrong. It turned out to parse fine on its own — but it is the fragile order, and that exposed a real bug.

`reorderFlagsFirst` glued the following token onto any flag without `=`, without knowing which flags take a value (`internal/cli/flagutil.go:71` on `main`). `--force` is boolean, so the stdlib parser consumes no value for it: it stopped at the glued skill name and silently discarded every flag after it.

- `omac register --force demo-echo --harness opencode` → usage dump, exit 2
- `omac register demo-echo --force --harness opencode` → `[ok] registered`

The reporter hit this by following omac's own advice: `skill-marketplace` exists under both opencode and claude-code, so the drift hint suggests `--force <skill>`, the ambiguity prompt then asks for `--harness`, and the combination dies. The bug is not register-specific — all 14 subcommands use this helper (reproduced on `deregister` too).

The display complaint is valid independently: `start.go:637` and `serve.go:989` printed a bare em dash with no label, so the line read as one run-on string, inconsistent with `register.go`'s disambiguation hints which were already labelled and skill-first.

## How

- `reorderFlagsFirst` takes the `*flag.FlagSet` and skips gluing for booleans (`internal/cli/flagutil.go:77`, `isBoolFlag` at `:91`). Unknown flags keep the value-taking behavior so typo diagnostics are unchanged. Threaded through all 14 call sites (one line each).
- New `internal/cli/hints.go`: `registerCmd` builds the command in documented order, `skillProblemLine` renders `<skill> — <label>: <cmd>` with the command colored. Keeps the four sites from drifting apart again.
- `CREATING_A_SKILL.md` + `oh-my-agentic-coder.md` updated and re-copied into the bundle (byte-identity enforced by `internal/builtinskills/builtinskills_test.go:113`).

Output before → after:

```
- demo-echo — omac register --force demo-echo
+ demo-echo — re-register: omac register demo-echo --force
```

## Verification

`go build ./...` — OK.

`go test ./internal/cli/ ./internal/builtinskills/` — both `ok`.

New tests:

- `TestReorderFlagsFirst_BoolFlagDoesNotSwallowPositional` — fails on `main`, pins the parser fix across three arg orders.
- `TestRegister_BoolFlagBeforeSkillKeepsLaterFlags` — the exact `--force <skill> --harness <h>` shape, and asserts the claude-code registry entry exists (i.e. `--harness` was honored, not dropped).
- `TestHintsUseDocumentedFlagOrder` — scans `internal/cli/*.go` and fails if any hint puts a flag before the skill name. Confirmed non-vacuous by temporarily reintroducing the old string.
- `TestRegisterCmd` / `TestSkillProblemLine` — pin the rendering.

Manual, built binaries from `main` and this branch (no setup needed; `nix-da` intentionally does not exist, so this isolates argument parsing):

```
$ omac register --force nix-da --harness claude
main:   Usage: omac register <skill> [flags]                            # --harness dropped
branch: omac register: "nix-da" not found in any in-scope skill source
```

Also verified the drift hint end-to-end by registering a skill, mutating its bundle and running `omac start`.

`go test ./...` — the two `internal/bridge` failures (`TestCodexBridge…`, `TestCopilotBridge…`) reproduce on unmodified `main` in this environment; they need node/bun, which is not installed. Everything else passes.

## Follow-up

The parser fix touches all 14 subcommands that reorder args, which is broader than the issue title suggests. Happy to split it into its own issue/PR if a reviewer prefers it isolated.

### DIFF
--- a/CREATING_A_SKILL.md
+++ b/CREATING_A_SKILL.md
@@ -637,7 +637,7 @@ checked:
    `node_modules/`, `.git/`, `.DS_Store`, `*.pyc`, editor swap files,
    so a `pip install` doesn't trip detection. Refuses unless
    `--accept-skill-changes` is passed (or you re-register with
-   `omac register --force my-skill`). The flag exists for the case
+   `omac register my-skill --force`). The flag exists for the case
    where you've intentionally edited the skill in place during
    development and don't want to re-prompt.
 
@@ -646,8 +646,10 @@ checked:
    resolvable `default_from_env:`). Refuses with the list of missing
    names and the exact register command:
    ```
-   omac start: my-skill: required config field(s) missing: API_BASE_URL, REGION
-   Run: omac register --reprompt-fields my-skill
+   omac start: refusing to start, found 1 problem(s):
+
+     required config field missing:
+       my-skill — fields: API_BASE_URL, REGION — set them with: omac register my-skill --reprompt-fields
    ```
 
 In all refusal cases the exit is non-zero so wrapper scripts can
@@ -771,7 +773,7 @@ Field names share the env-var namespace with `secrets:` and
 Two ways:
 
 ```bash
-omac register --reprompt-fields my-skill   # interactive, keeps secrets
+omac register my-skill --reprompt-fields   # interactive, keeps secrets
 ```
 
 …or edit `.opencode/skill-config.yaml` directly (mode 0600 — chmod it
--- a/internal/builtinskills/assets/omac-write-a-skill/references/creating-a-skill.md
+++ b/internal/builtinskills/assets/omac-write-a-skill/references/creating-a-skill.md
@@ -637,7 +637,7 @@ checked:
    `node_modules/`, `.git/`, `.DS_Store`, `*.pyc`, editor swap files,
    so a `pip install` doesn't trip detection. Refuses unless
    `--accept-skill-changes` is passed (or you re-register with
-   `omac register --force my-skill`). The flag exists for the case
+   `omac register my-skill --force`). The flag exists for the case
    where you've intentionally edited the skill in place during
    development and don't want to re-prompt.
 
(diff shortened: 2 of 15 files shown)


nhuelstng:
Parser fix looks right, reproduced the bug on `main` and it's gone here. But it also changes inner-arg routing in `start`/`serve`:

```
omac start opencode --verbose run "fix it" --model x
main:   forwards [run --model x] to opencode
branch: exit 2, flag provided but not defined: -model
```

`start.go:146` and `serve.go:110` do `innerArgs = append(fs.Args(), innerArgs...)`, so in those two the early stop was routing the later flags to the harness, not dropping them. The other 12 validate `fs.Args()`, which is where the #227 loss actually happened.

I'm fine with the new behavior — `--` is the documented channel and we share `--verbose`/`-s` with the harnesses. Just make it deliberate: mention it in the commit, cover `start`/`serve` with a test, and point at `--` when parse fails there instead of blaming omac for opencode's flag.

willm-TNG:
Thanks for pointing that out. Kept the new start/serve routing. -- is the pass-through. Follow-up commit documents that, pins both the -- form and the mixed form for start and serve, and on parse failure prints the -- reminder before usage instead of leaving it as an unknown omac flag.

VictoriaRuckerbauer (approved):

nhuelstng (approved):

PR #230, opened 2026-08-13
fix(sandbox,cli): grant every hop of the harness symlink chain

### LINKED ISSUE #229: macOS: harness reached through a multi-hop symlink chain fails with a misleading ENOENT
## Context / Summary

Reported on omac 0.7.1 (macOS 26.6.1, arm64). `omac start opencode -- --version`
exits 71 with

```
sandbox-exec: execvp() of 'opencode' failed: No such file or directory
```

on a machine where `which opencode` resolves fine. The same binary launched by
its fully resolved path (`omac start opencode -inner <realpath> -- --version`)
runs and prints its version. So the harness is present and executable — what
fails is reaching it from inside the sandbox.

## Problem / What

The reporter's distinguishing property is Homebrew in a **non-default prefix**
(`~/homebrew`), which no baseline grant covers —
`internal/sandboxprofile/baseline.go:106` grants `/opt/homebrew`,
`/usr/local/Cellar`, `/usr/local/opt` only. The harness is then reached through
a **three-hop** symlink chain:

```
~/homebrew/bin/opencode
  -> ~/homebrew/Cellar/opencode/1.18.15/bin/opencode                          <- never granted
  -> ~/homebrew/Cellar/opencode/1.18.15/libexec/.../opencode-ai/bin/opencode.exe
```

`resolveCommandBinaryDirs` (`internal/sandboxrun/resolve_binary.go:47`) grants
only the two **ends** of that chain — `Dir(PATH entry)` and
`Dir(EvalSymlinks(...))`. The kernel resolves a chain one hop at a time and
needs read access to each link it reads on the way, so resolution dies on the
middle directory. Seatbelt reports the denied in-sandbox PATH lookup of a bare
`argv[0]` as ENOENT, which names neither the deny nor the path.

Isolated from omac with a hand-written SBPL granting exactly the two ends —
three spellings of the *same* file behave differently, which is what pins the
cause to path resolution rather than to exec:

| `argv[0]` | result |
| --- | --- |
| `opencode` (bare — what omac passes) | `execvp() ... No such file or directory`, rc 71 |
| `/…/homebrew/bin/opencode` (absolute, still the link) | `Operation not permitted` |
| `/…/libexec/…/opencode.exe` (resolved) | runs, rc 0 |

Adding a read grant on the middle directory makes all three rc 0.

Two secondary gaps make the failure less readable than it needs to be:

- `BuildChildArgv` hands `sandbox-exec` a bare `argv[0]`
  (`internal/sandboxrun/backend_darwin.go:39`), so the PATH lookup happens a
  second time *inside* the sandbox, where a failure surfaces as ENOENT.
- The pre-flight checks `harness.InnerCmd` rather than the resolved argv
  (`internal/cli/start.go:250`), so a profile-pinned `inner_cmd` validates the
  wrong binary and `--inner` skips the check entirely.

## Suggested fix / Ask

- [ ] Grant the directory of **every hop** of the symlink chain, not just the two ends
- [ ] Same for interpreter resolution — nvm/mise/asdf shims are multi-hop too
- [ ] Pass `argv[0]` to `sandbox-exec` as an absolute path so the second, in-sandbox lookup disappears
- [ ] Pre-flight the *resolved* argv, and warn up front when the inner command is unresolvable or sits inside a protected-path deny

### Acceptance criteria

- [ ] `omac start opencode -- --version` succeeds against a `$HOME`-rooted Homebrew prefix with a three-hop chain and no profile grant for that prefix
- [ ] A profile that denies the tree holding the harness still denies it, and says so before the launch fails
- [ ] Unit coverage for the multi-hop chain, for cycle termination, and for the unchanged single-hop/plain-binary cases

## Non-goals

- **Resolving the harness out of a denied tree.** Exec of a main image consults
  `(allow process-exec*)` and never the file-read rules, so passing the
  `EvalSymlinks` form would launch a binary the user's profile explicitly
  denies. omac must report that conflict, not route around it.
- Widening the baseline to cover `$HOME`-rooted install prefixes wholesale.

## Evidence / Environment

- omac 0.7.1, macOS 26.6.1 arm64, Homebrew prefix `~/homebrew`, opencode 1.18.15
- Repro: fixture `$HOME` with the three-hop chain, profile with neither a
  `filesystem.deny` nor a grant for the prefix. The fixture must live **outside
  `/tmp`** — the baseline grants `/tmp` and `/private/tmp`, which masks the bug
  entirely.
- 0.7.1 → exit 71 with the ENOENT above; a build carrying the chain grant →
  prints the version, exit 0.
- Unrelated to this issue: the reporter's log also shows
  `sandbox profile "builtin" could not be resolved ... GET /sandbox/denied disabled`.
  That is already fixed on `main` by #219 and is absent from current builds.

NoRiceToday:
**Issue:** Closes #229

## What

- `resolveCommandBinaryDirs` walks the symlink chain and grants the directory of every hop, not just the two ends (`symlinkChainDirs`, `internal/sandboxrun/resolve_binary.go`)
- Same walk for interpreter resolution, so multi-hop nvm/mise/asdf shims stop failing the exec of every script naming them
- `BuildChildArgv` hands `sandbox-exec` an absolute `argv[0]`, including the wrapped command of an `env NAME=VALUE …` prefix
- The pre-flight validates the *resolved* argv; two non-fatal warnings name the cause before the launch dies inside Seatbelt

## Why

Bug report + full diagnosis in #229. Short version: the middle directory of a three-hop chain was never granted, so path resolution died there and macOS reported it as ENOENT on a binary that is plainly on `$PATH`.

## How

- The chain walk is `Lstat`/`Readlink` per hop, capped at 32 hops so a cycle terminates, plus the `EvalSymlinks` spelling of the end for symlinked ancestor dirs. Directories are granted, not links — a subpath grant is what lets the kernel read the link and stat its siblings, matching how the rest of the file grants access.
- `lookPathAbs` deliberately does **not** follow symlinks. Under a deny covering the harness's tree all three `argv[0]` forms were measured, and only the `EvalSymlinks`-resolved one runs — because exec of a main image consults `(allow process-exec*)` and never the file-read rules. Substituting it would launch a binary out of a tree the user's profile denies, so `protectedInnerBinaryWarning` reports the conflict instead. Grant order is unchanged (read allows → protected denies → write allows), so denies still win over the new grants.
- Absolute `argv[0]` is not redundant with the chain grant: it also makes the launch independent of the child's `PATH`, which `deny_vars` can strip since it is applied after `BaseAllowVars`.

## Verification

Fixture reproducing the report: `$HOME` outside `/tmp` (a `/tmp`-rooted one is masked by the baseline `/tmp` grant), Homebrew prefix at `$HOME/homebrew`, three-hop chain to the real opencode 1.18.6 binary, profile with neither a `filesystem.deny` nor a grant for the prefix.

```
$ omac-v0.7.1 start opencode -- --version
sandbox-exec: execvp() of 'opencode' failed: No such file or directory   exit 71
$ omac-this-branch start opencode -- --version
1.18.6                                                                   exit 0
```

Kernel behaviour isolated from omac with a hand-written SBPL granting exactly the two ends of the chain — the three-form table in #229. Adding the middle directory takes all three forms to rc 0.

```
$ gofmt -l internal/ cmd/     # clean
$ go build ./...              # ok
$ go test ./...               # EXIT=0, all packages ok
$ go test ./internal/sandboxrun/ -run 'SymlinkHop|SymlinkChainDirs' -v
--- PASS: TestResolveInnerBinaryDirsGrantsEverySymlinkHop
--- PASS: TestResolveInnerBinaryDirsGrantsEverySymlinkHopViaPATH
--- PASS: TestSymlinkChainDirsTerminatesOnCycle
--- PASS: TestSymlinkChainDirsPlainBinary
```

macOS only, by hand — the Linux backend execs the resolved path directly, so it never traversed the chain, but it shares `resolveCommandBinaryDirs` and picks the wider ro-bind set up.

## Follow-up

The report's log also carried `sandbox profile "builtin" could not be resolved … GET /sandbox/denied disabled`. Unrelated, already fixed on `main` by #219, absent from this build — noted here only so it is not re-reported off the same transcript.

### DIFF
--- a/internal/cli/start.go
+++ b/internal/cli/start.go
@@ -172,15 +172,25 @@ func parseLaunchArgs(cmdName string, args []string, env *Env) (launchOpts, int) 
 	}, ExitOK
 }
 
-// checkInnerBinary verifies the harness's inner command binary is on $PATH.
+// checkInnerBinary verifies the RESOLVED inner command binary is on $PATH.
 // Returns ExitOK when found, ExitPrerequisiteMissing when missing, ExitOK
-// when InnerCmd is empty (defensive skip). Called by runLaunch and runServe.
-func checkInnerBinary(harness config.Harness, prefix string, env *Env) int {
-	if len(harness.InnerCmd) == 0 {
+// when innerCmd is empty (defensive skip). Called by runLaunch and runServe.
+//
+// It takes the resolved argv — Harness.ResolveInnerCmd's output — rather than
+// a Harness, because that argv is what the sandbox backend will actually try
+// to resolve. Checking only the harness default left two routes unguarded: a
+// profile-pinned inner_cmd (reached when the harness has no default of its
+// own) and an explicit --inner override, whose call sites skipped the
+// pre-flight outright. Either then fails much further downstream, where the
+// backend's own PATH lookup silently yields no grant and macOS reports the
+// denied in-sandbox lookup as a bare "No such file or directory" that names
+// the wrong cause.
+func checkInnerBinary(innerCmd []string, prefix string, env *Env) int {
+	if len(innerCmd) == 0 || innerCmd[0] == "" {
 		return ExitOK
 	}
-	if _, err := exec.LookPath(harness.InnerCmd[0]); err != nil {
-		fmt.Fprintf(env.Stderr, "%s: harness binary %q not found on $PATH; install it or pass --inner-cmd <path>\n", prefix, harness.InnerCmd[0])
+	if _, err := exec.LookPath(innerCmd[0]); err != nil {
+		fmt.Fprintf(env.Stderr, "%s: harness binary %q not found on $PATH; install it or pass --inner-cmd <path>\n", prefix, innerCmd[0])
 		return ExitPrerequisiteMissing
 	}
 	return ExitOK
@@ -247,9 +257,17 @@ func runLaunch(env *Env, opts launchOpts) int {
 	profName := plan.Name
 	prof := plan.Launcher
 
-	// 1b. Pre-flight: inner harness binary must be on $PATH.
+	// 1b. Pre-flight: inner harness binary must be on $PATH. Checked on the
+	//     RESOLVED argv — profile inner_cmd, else the harness default — which
+	//     is the same argv step 8 hands to the sandbox. Passing the harness
+	//     default instead validated the wrong binary whenever a config pinned
+	//     inner_cmd: it confirmed opencode while the launch ran something
+	//     else, which then failed inside Seatbelt naming the wrong cause.
+	//     An explicit --inner still skips: pointing at an exact binary is a
+	//     deliberate escape hatch. sandboxrun warns non-fatally if it cannot
+	//     be resolved either.
 	if innerCmdOverride == "" {
-		if code := checkInnerBinary(harness, prefix, env); code != ExitOK {
+		if code := checkInnerBinary(harness.ResolveInnerCmd(prof.InnerCmd, ""), prefix, env); code != ExitOK {
 			return code
 		}
 	}
--- a/internal/cli/start.go
+++ b/internal/cli/start.go
@@ -22,6 +22,7 @@ import (
 	"github.com/tngtech/oh-my-agentic-coder/internal/keychain"
 	"github.com/tngtech/oh-my-agentic-coder/internal/registry"
 	"github.com/tngtech/oh-my-agentic-coder/internal/sandbox"
+	"github.com/tngtech/oh-my-agentic-coder/internal/sandboxrun"
 	"github.com/tngtech/oh-my-agentic-coder/internal/secrets"
 	"github.com/tngtech/oh-my-agentic-coder/internal/session"
 	"github.com/tngtech/oh-my-agentic-coder/internal/skillconfig"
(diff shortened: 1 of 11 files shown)


internal/cli/start.go
```diff
@@ -165,15 +165,25 @@ func parseLaunchArgs(cmdName string, args []string, env *Env) (launchOpts, bool)
...
+// backend's own PATH lookup silently yields no grant and macOS reports the
+// denied in-sandbox lookup as a bare "No such file or directory" that names
+// the wrong cause.
+func checkInnerBinary(innerCmd []string, prefix string, env *Env) int {
+	if len(innerCmd) == 0 || innerCmd[0] == "" {
 		return ExitOK
 	}
-	if _, err := exec.LookPath(harness.InnerCmd[0]); err != nil {
-		fmt.Fprintf(env.Stderr, "%s: harness binary %q not found on $PATH; install it or pass --inner-cmd <path>\n", prefix, harness.InnerCmd[0])
+	if _, err := exec.LookPath(innerCmd[0]); err != nil {
```
willm-TNG: Non-blocking — latent today, since no shipped config can reach it.

**Preflight can pass while the harness is missing.**

When a profile pins an `env` wrapper, the resolved argv looks like:

    ["env", "NPM_CONFIG_CACHE=/x", "opencode", "--version"]

`innerCmd[0]` is then `env`, not `opencode`. `/usr/bin/env` always resolves,
so `checkInnerBinary` returns `ExitOK` even when `opencode` isn't installed —
exactly the late, mislabeled failure this PR set out to prevent.

`sandboxrun` already solves this: `resolveInnerBinaryDirs`, `absoluteInnerArgv`
and `unresolvedInnerCommandWarning` all skip the wrapper via
`unwrapEnv`/`envWrapperEnd`. Suggest doing the same here rather than adding a
second copy of that logic in `cli` — which means exporting `unwrapEnv`, since
it's unexported today.

One wrinkle: `unwrapEnv` skips `NAME=VALUE` tokens but intentionally leaves
`env` flags alone, so `env -i opencode` unwraps to `["-i", "opencode"]`. The
lookup should skip a leading `-` token instead of reporting
`harness binary "-i" not found on $PATH`.

On severity: every shipped harness default is a single plain binary and the
sandboxed profiles ship an empty `inner_cmd`, so this needs a hand-written
`inner_cmd` with an `env` prefix plus a missing harness. Even then the built-in
profiles still warn downstream via `unresolvedInnerCommandWarning`; a custom
profile that doesn't route through `omac sandbox run` is the case that loses
the diagnostic entirely.

willm-TNG (approved):

PR #234, opened 2026-08-14
fix(config,cli,skillsource): honor a relocated harness config home

### LINKED ISSUE #233: A relocated harness config home (CLAUDE_CONFIG_DIR) is silently ignored on both sides of the sandbox
### Duplicate check

### Context / Summary

Users keep separate logins side by side by relocating a harness's config home —
for Claude Code that's `CLAUDE_CONFIG_DIR` (e.g. `~/.claude` personal,
`~/.work-claude` work). omac models this per harness as `Harness.HomeEnv`, but
claude's is declared as `CLAUDE_HOME`, a name nothing reads. So the redirect is
ignored on both sides of the sandbox boundary: omac resolves every path from the
default home, and the sandboxed harness is never told the redirect exists. The
user silently gets the default config home instead of the one they selected.

A patch is ready and will be raised as a PR against this issue.

### Problem / What

- `internal/config/harness.go:281` declares claude's `HomeEnv` as `CLAUDE_HOME`
  — nothing reads that name. Claude Code's own variable is `CLAUDE_CONFIG_DIR`,
  so `ConfigHome()` returns `~/.claude` whatever the user set, and with it the
  session store, `omac setup`'s install dir, and skill discovery.
- The redirect never reaches the sandboxed harness. `internal/cli/serve.go:713`
  forwards only `harness.SandboxEnvAllow`, no harness lists its own `HomeEnv`,
  and `sandboxprofile.DefaultAllowVars()` doesn't either — so `FilterEnv` strips
  it and the harness falls back to `~/.claude`. **The failure is silent:** with
  both homes present the user gets the wrong login; with only the redirect
  populated, `claude` finds no credentials in `~/.claude` and prompts a fresh
  login.
- The one profile shape where the var *does* survive is the sanctioned
  `allow_vars: ["*"]` opt-in — and there it breaks the other way: the harness
  reads the relocated home, which `internal/cli/start.go:926` and
  `internal/cli/serve.go:624` never grant (they inject `harness.SandboxDirs`
  verbatim, declared against the default home), so it's denied.
- Skill discovery derives its user-global roots from `$HOME` and the skills base
  alone (`internal/skillsource/skillsource.go:118`), so it cannot follow a
  relocated config home even once the above is fixed — while `omac setup`
  installs into `GlobalSkillsDir()` (`internal/cli/setup.go:63`, `:104`), which
  does follow it. Install target and discovery would then disagree.
- Same root cause, visible without any redirect: pi's skills live in
  `~/.pi/agent/skills` and codewhale's in `~/.codewhale/skills`, neither
  derivable from the skills bases, yet both are `omac setup` targets — so
  built-in skills are installed where discovery never looks.

### Suggested fix / Ask

Make the config-home override authoritative everywhere omac derives a path from
it.

- [ ] Point claude's `HomeEnv` at `CLAUDE_CONFIG_DIR` (the variable Claude Code
      reads).
- [ ] Forward `HomeEnv` into the sandbox generically for every harness that
      declares one — not per descriptor, so it cannot be declared for one
      harness and forgotten for the next.
- [ ] Resolve the grants: swap the `SandboxDirs` entry naming the default config
      home for the resolved one, so the redirect target is granted and the
      default one is not (granting both would defeat the separation the user
      asked for). This must land *with* the forwarding above — swapping grants
      while the harness still can't see its `HomeEnv` is worse than not
      swapping, since omac would grant one directory while the harness reads
      another.
- [ ] Handle the harness whose config home is not itself a declared sandbox dir
      (pi declares `~/.pi`, its config home is `~/.pi/agent`) — nothing to swap,
      so the redirect target must still be granted.
- [ ] Normalize the override value (tilde-expand, clean, absolutize) before
      comparing it to the default, so a value that merely spells the default
      home differently isn't treated as a redirect.
- [ ] Make skill discovery scan the harness's own `GlobalSkillsDir()`,
      substituting the redirected root for the default one *in place* so a
      redirect doesn't reorder the rest of the candidate ladder. This also
      closes the pi/codewhale gap above.
- [ ] Neutralize ambient `HomeEnv` vars in tests. Tests fake `$HOME`, but an
      override names an absolute path and survives that, so a developer who
      genuinely uses `CLAUDE_CONFIG_DIR` gets failures CI never sees.
- [ ] Document `CLAUDE_CONFIG_DIR` support in `docs/CLI.md`, including that a
      redirect hides skills installed under the default home.

## Acceptance criteria

- [ ] `claude` launched via omac with `CLAUDE_CONFIG_DIR` set uses the login from
      *that* home — not the default home's, and not a fresh login prompt.
- [ ] Skills installed under the relocated home are discovered; skills under the
      superseded default home are not.
- [ ] Every harness declaring a `HomeEnv` forwards it, asserted by a
      registry-wide test rather than one test per harness.
- [ ] Setting `HomeEnv` to a non-canonical spelling of the default home (trailing
      slash, `..`, `~`) changes nothing.
- [ ] `go test ./...` shows no new failures; e2e matrix green on Linux and macOS.

### Non-goals

- Auto-forwarding provider-auth env vars for harnesses that deliberately don't
  declare them (opencode, pi, codewhale). `HomeEnv` is a path, not a credential.
- Migrating or aliasing the old `CLAUDE_HOME` name. It was undocumented and
  unread — nothing depended on it.
- Changing how skills are *installed*; only where they are discovered from.

### Evidence / Environment

Reproduces on `main` (Linux, `go1.26.6`). Probe with `HOME=/home/u` and
`CLAUDE_CONFIG_DIR=/home/u/.work-claude`, using `Harness` plus
`sandboxprofile.FilterEnv`:

```
declared HomeEnv            = "CLAUDE_HOME"              # user's CLAUDE_CONFIG_DIR is never read
ConfigHome()                = "/home/u/.claude"          # not /home/u/.work-claude
GlobalSkillsDir()           = "/home/u/.claude/skills"   # omac setup installs into the default home
SandboxDirs (granted as-is) = [~/.claude ~/.local/share/claude]
SandboxEnvAllow             = [ANTHROPIC_API_KEY ANTHROPIC_AUTH_TOKEN ANTHROPIC_BASE_URL ANTHROPIC_MODEL]

CLAUDE_CONFIG_DIR reaches the harness?
  default profile allow_vars -> false   # stripped: harness silently uses ~/.claude -> potentially logged in with the wrong account
  allow_vars ["*"]           -> true    # passes: harness reads ~/.work-claude, which is not granted -> asks for login
```

That `CLAUDE_CONFIG_DIR` is the name Claude Code actually reads is confirmed
empirically: a patched build honoring it reuses the relocated home's existing
login (manually verified on Linux).

NiklasEi:
**Issue:** Closes #233 

## What
- Claude Code's config-home override now works: `HomeEnv` was declared as
  `CLAUDE_HOME`, a name nothing reads (`CLAUDE_CONFIG_DIR` is the real one).
- Every harness's `HomeEnv` is forwarded into the sandbox, and the sandbox grants
  follow the resolved config home instead of the default one.
- Global skill discovery follows a relocated config home — and now also finds
  pi's and codewhale's config-home skills dirs, which it never scanned.
- OpenCode's `HomeEnv` is **removed**: `OPENCODE_HOME` was omac's invention too,
  and OpenCode has no config-home override to honor.

## Why
Users relocate a harness's config home to keep separate logins side by side
(`~/.claude` personal, `~/.work-claude` work). omac claimed to model this but
honored it nowhere the harness could see, so the redirect was **silently
ignored**: omac resolved every path from the default home, and the sandboxed
harness was never told the redirect existed. With both homes present the user
gets the wrong login; with only the redirect populated, `claude` finds no
credentials in `~/.claude` and prompts a fresh login.

The one profile shape where the var survived — the sanctioned
`allow_vars: ["*"]` — broke the other way: the harness read the relocated home,
which was never granted, so the sandbox denied it.

## How
1. **`5010e06` — config + launch path.** `ConfigHome()` normalizes the value, so a
   different *spelling* of the default home is not read as a redirect.
   `ResolvedSandboxDirs()` swaps the `SandboxDirs` entry naming the default config
   home — appending instead when no entry names it (pi declares `~/.pi`, its home
   is `~/.pi/agent`). `ForwardedEnvVars()` adds `HomeEnv` generically
   (`internal/cli/serve.go:713`); that forwarding is what makes the swap sound,
   since granting the relocated home while the harness reads the default one is
   worse than not swapping at all.
2. **`fc94e08` — discovery.** `withConfigHomeRoot` substitutes the redirected
   skills root for the default one *in place*, so a redirect moves which config
   home is scanned without reordering the rest of the ladder. It also makes
   `GlobalSkillsDir()` always a candidate, closing a pre-existing gap: pi and
   codewhale are `omac setup` targets discovery never looked in.
3. **`b5d2a45` — opencode.** `OPENCODE_CONFIG_DIR` only *adds* a config-search dir
   ([docs](https://opencode.ai/docs/config#custom-directory)) and credentials live
   outside it, so it is no `HomeEnv`: declaring it would swap `~/.config/opencode`
   out of the grants while OpenCode kept reading it. The field doc now carries the
   general rule — no upstream variable ⇒ leave `HomeEnv` empty.

Test isolation came along by necessity: tests fake `$HOME`, but an override names
an absolute path and survives that, so an ambient `CLAUDE_CONFIG_DIR` pointed them
at the developer's real config home — 9 failures locally, 0 in CI. A `TestMain`
per package clears them via the new `config.HomeEnvNames()`.

## Verification
Linux, `go1.26.6`. Every CI lint gate, run locally:

```
gofmt -l .                               # clean
go vet ./...                             # clean
staticcheck ./...                        # clean
python3 scripts/check-docs.py            # ok
python3 scripts/check-workflow-shell.py  # ok
go test ./...                            # same failure set as main
```

- **No regressions:** `go test ./...` fails the same 51 tests as `main` —
  pre-existing loopback/network artifacts of the local sandbox in `cli`, `facade`,
  `intent`, `netprompt/origin`, `netproxy`, `sandboxrun`, `updater` — confirmed by
  diffing both failure lists.
- **Manual:** a build with `CLAUDE_CONFIG_DIR` set reuses that home's login
  instead of prompting — which also confirms it is the name Claude Code reads.
- **Not run:** `go mod verify` (blocked on `proxy.golang.org` locally) and the
  e2e matrix. This touches launch argv and macOS/Seatbelt is untested, so
  `e2e.yml` is the gate.

### Every `HomeEnv` name audited

| Harness | Declared | Evidence |
|---|---|---|
| claude-code | `CLAUDE_CONFIG_DIR` | 147 occurrences in installed claude 2.1.228 (`CLAUDE_HOME`: 0) |
| opencode | *(none, this PR)* | `OPENCODE_HOME`: 0 occurrences; `OPENCODE_CONFIG_DIR` is additive only |
| pi | `PI_CODING_AGENT_DIR` | 3 occurrences in `pi-coding-agent/dist/cli.js` |
| codex | `CODEX_HOME` | `openai/codex`, `codex-rs/utils/home-dir/src/lib.rs`: `find_codex_home()` — "the Codex configuration directory, which can be specified by the `CODEX_HOME` environment variable […] defaults to `~/.codex`" |
| copilot | `COPILOT_HOME` | `github/copilot-cli` changelog: "`--config-dir` is deprecated in favor of `COPILOT_HOME`" |
| codewhale | `CODEWHALE_HOME` | `Hmbown/CodeWhale`, `crates/paths/src/lib.rs`: `codewhale_home_override()`; `codewhale_home_is_explicit()` documents it as an "explicit isolation boundary" |

Three upstream divergences surfaced by the audit — all narrow, none blocking:

- **codex canonicalizes** `CODEX_HOME` (resolves symlinks) and errors if it does
  not exist; `normalizeHomePath` only cleans and absolutizes. A symlinked value
  means omac grants the link path while codex uses the target.
- **codewhale rejects** a relative `CODEWHALE_HOME`; omac absolutizes it against
  its own working directory instead.
- **copilot stops loading `~/.agents/skills`** when `COPILOT_HOME` is set, but omac
  keeps the shared `agents` root in copilot's scope — the same discovery/harness
  mismatch this PR fixes for claude, in a spot it does not reach.

## Follow-up
- **Document it.** `docs/CONFIGURATION.md` has no harness-config-home section —
  `CLAUDE_CONFIG_DIR` and the per-harness overrides belong there, together with
  the behavior change that a redirect **hides** skills installed under the
  default home. Worth a `docs/` commit on this branch or a separate PR.
- **Symlinked / relative override values.** `normalizeHomePath` cleans and
  absolutizes but does not resolve symlinks, so it can disagree with codex
  (canonicalizes) and codewhale (rejects relative values). Worth a follow-up issue
  rather than widening this PR.
- **copilot + the shared `agents` skills root.** `COPILOT_HOME` makes copilot stop
  reading `~/.agents/skills` while omac still treats it as in scope for copilot.

🤖 Generated with Claude Opus 5

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -52,6 +52,23 @@ The project file **replaces** the global one; omac does not merge the two. Any o
 
 The launcher config changes only *how omac launches* in the project: the sandbox runtime it selects, the cache scope, and the facade and audit settings. It does **not** change what the agent is allowed to access. Those grants (filesystem paths, network hosts, open ports) come from the user-global sandbox grants file described below and **currently have no per-project equivalent**.
 
+## Harness config home
+
+Each harness can relocate its configuration directory (its "config home") via an environment variable — typically to keep two logins side by side, for example a personal and a work `~/.claude`. omac follows the redirect: the redirected directory is granted to the sandbox instead of the default one and created if it does not exist yet, the variable is passed through to the harness, and global skills and resumable sessions are read from the redirected directory.
+
+| Harness | Variable | Default |
+|---|---|---|
+| claude-code | `CLAUDE_CONFIG_DIR` | `~/.claude` |
+| codex | `CODEX_HOME` | `~/.codex` |
+| copilot | `COPILOT_HOME` | `~/.copilot` |
+| pi | `PI_CODING_AGENT_DIR` | `~/.pi/agent` |
+| codewhale | `CODEWHALE_HOME` | `~/.codewhale` |
+| opencode | *(none)* — relocate with `$XDG_CONFIG_HOME` | `~/.config/opencode` |
+
+Example: `CLAUDE_CONFIG_DIR=~/.work-claude omac start claude` runs Claude Code with the work login instead of prompting for a fresh one.
+
+A redirect also **hides** the skills installed under the default home: omac scans only the redirected directory for global skills. After switching homes, run `omac setup` to provision the built-in skills there too.
+
 ## Sandbox grants
 
 The sandbox grants file (`~/.config/omac/sandbox-profiles/default.json`) controls what the agent is actually allowed to access — filesystem paths, network mode, and environment variables. This is separate from the launcher config above, which only selects which sandboxing technology to use.
--- a/internal/cli/continue_resume_test.go
+++ b/internal/cli/continue_resume_test.go
@@ -345,6 +345,28 @@ func TestLaunchOpenCodeCreatesAndGrantsRuntimeDirs(t *testing.T) {
 	t.Fatalf("start argv missing opentui read/write grant: %v", capture.args)
 }
 
+// A redirect at a not-yet-existing home must be created before grant
+// resolution, or its read+write grant is dropped.
+func TestLaunchCreatesRedirectedConfigHome(t *testing.T) {
+	// The target deliberately does not exist yet; it sits outside the
+	// helper's faked $HOME, whose path the helper picks itself.
+	base := t.TempDir()
+	redirected := filepath.Join(base, ".work-claude")
+	t.Setenv("CLAUDE_CONFIG_DIR", redirected)
+
+	capture := launchCacheCaptureForHarness(t, "claude", false, false, false)
+	if info, err := os.Stat(redirected); err != nil || !info.IsDir() {
+		t.Fatalf("redirected config home was not created before start: info=%v err=%v", info, err)
+	}
+
+	for i := 0; i+1 < len(capture.args); i++ {
+		if capture.args[i] == "--allow" && capture.args[i+1] == redirected {
+			return
+		}
+	}
+	t.Fatalf("start argv missing redirected config home grant: %v", capture.args)
+}
+
 func TestLaunchCacheNoSandboxPreservesHostEnvironment(t *testing.T) {
 	capture := launchCacheCapture(t, true, false, false)
 	for key := range toolcache.Environment("ignored", toolcache.ModePersistent) {
--- a/internal/cli/continue_resume_test.go
+++ b/internal/cli/continue_resume_test.go
@@ -693,7 +693,7 @@ func writeClaudeSessionFile(t *testing.T, home, workdir, id, ts string) {
 // created — not the sibling.
 func TestContinueHintSkipsSiblingSession(t *testing.T) {
 	home := t.TempDir()
-	t.Setenv("CLAUDE_HOME", home)
+	t.Setenv("CLAUDE_CONFIG_DIR", home)
 	h, ok := config.LookupHarness("claude-code")
 	if !ok {
 		t.Fatal("claude-code harness not registered")
--- a/internal/cli/main_test.go
+++ b/internal/cli/main_test.go
@@ -5,6 +5,8 @@ import (
 	"testing"
 
 	"github.com/zalando/go-keyring"
+
+	"github.com/tngtech/oh-my-agentic-coder/internal/config"
 )
 
 // TestMain installs go-keyring's in-memory mock provider for the whole
@@ -17,7 +19,20 @@ import (
 //
 // The mock returns ErrNotFound for absent keys and stores Set values in
 // memory, which is exactly the deterministic behavior these tests assume.
+//
+// It also clears every harness config-home override (CLAUDE_CONFIG_DIR and
+// friends). Tests here fake $HOME but cannot fake a variable that names an
+// absolute path, so an ambient override kept pointing at the developer's real
+// config home.
 func TestMain(m *testing.M) {
 	keyring.MockInit()
+	clearHarnessHomeEnvs()
 	os.Exit(m.Run())
+}
+
+// clearHarnessHomeEnvs unsets every harness HomeEnv for the whole test binary.
+func clearHarnessHomeEnvs() {
+	for _, name := range config.HomeEnvNames() {
+		os.Unsetenv(name)
+	}
 }
--- a/internal/cli/main_test.go
+++ b/internal/cli/main_test.go
@@ -21,16 +21,14 @@ import (
 // memory, which is exactly the deterministic behavior these tests assume.
 //
 // It also clears every harness config-home override (CLAUDE_CONFIG_DIR and
-// friends). Tests here fake $HOME but cannot fake a variable that names an
-// absolute path, so an ambient override kept pointing at the developer's real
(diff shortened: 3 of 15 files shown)


nhuelstng (approved):

mwtng (approved):
I rebased on the new main and added 3 commits:
- streamline comments for easier reading
- fix-up of imports for new file, create redirected home if non-existent before grant resolution
- add documentation

nhuelstng (approved):

PR #248, opened 2026-08-19
fix(config,skilltrust): resolve symlinked skill roots in BundleHash a…

### LINKED ISSUE #247: BundleHash returns the empty hash for a skill directory that is a symlink
### Duplicate check

### Context / Summary

I have the opencode skills folder as a symlink to a git-versioned repo (`skills -> $AI-Config/opencode/skills/`). Registering such a skill records bundle_hash: sha256:e3b0c442… — the SHA-256 of empty input — and an empty approval snapshot, so `omac start` spawns the sidecar from an empty directory and the health check fails with connection refused.

### Problem / What

config.BundleHash (internal/config/meta.go:359) resolves the skill dir with filepath.Abs then walks it with filepath.WalkDir. WalkDir Lstats the root; when the root is a symlink it reports ModeSymlink, the walk callback fires once with rel == "." (skipped), and the walk does not descend. Result: zero items → sha256:e3b0c44….

This is inconsistent with skilltrust.copyTree (internal/skilltrust/snapshot.go:135), which EvalSymlinks the root first and copies the real tree — so the recorded hash (empty) and the snapshot (real files, when present) disagree, and IsApproved(name, realHash) never matches.

### Suggested fix / Ask

- [ ] In config.BundleHash, EvalSymlinks the resolved abs before walking (mirror copyTree). On a not-exist error, fall through to WalkDir so the missing-dir contract (TestBundleHash_MissingDir) still gets its error from the walk, not from EvalSymlinks.
- [ ] Add TestBundleHash_SymlinkRoot in internal/config/bundle_test.go: stage a real dir, symlink it, assert BundleHash(symlink) == BundleHash(real) and that the hash differs from sha256:e3b0c44….
- [ ] Add a skilltrust test that Approve against a symlinked skill dir produces a non-empty snapshot whose contents match the real tree (guards the copyTree/BundleHash consistency).
- [ ] Note in docs/SECURITY_MODEL.md (where config.BundleHash is already referenced, :129) that skill roots may be symlinks and both hash and snapshot resolve them.

### Non-goals

- Changing in-tree symlink handling (still dropped, by design — internal/config/meta.go:392).
- Changing the bundle-hash wire format.

### Evidence / Environment

- omac 0.7.1, Linux.
- readlink ~/.config/opencode/skills/skill-marketplace → ../../library/skills/skill-marketplace/.
- Repro: omac register --global skill-marketplace against the symlink, then omac start.

BenediktBurger:
…nd copyTree

**Issue:** Closes #247 

## What
Resolve symlinks of skill root folders before hashing.

## Why
See #247

## How
- resolve root symlinks in `BundleHash`.

## Verification
~~~
go test ./internal/config/ ./internal/skilltrust/

ok  	github.com/tngtech/oh-my-agentic-coder/internal/config	(cached)
ok  	github.com/tngtech/oh-my-agentic-coder/internal/skilltrust	(cached)
~~~

### DIFF
--- a/docs/SECURITY_MODEL.md
+++ b/docs/SECURITY_MODEL.md
@@ -128,7 +128,17 @@ the sandbox:
 
 - `config.BundleHash` deliberately excludes dependency/artifact subtrees
   (`node_modules`, `.venv`, `build`, `dist`, `target`, …) and does not follow
-  symlinks, so the hash alone does not cover everything a skill executes.
+  in-tree symlinks, so the hash alone does not cover everything a skill executes.
+
+  The skill root itself may be a symlink — either the skills directory
+  (~/.config/opencode/skills -> /repo/opencode/skills/) or an individual
+  skill within it (skills/foo -> ../../library/skills/foo/), common for
+  git-versioned skill libraries. Both `config.BundleHash` and the approval
+  snapshot (`skilltrust.copyTree`) resolve the root via
+  `filepath.EvalSymlinks` before walking, so they agree on what they
+  cover, while in-tree symlinks remain dropped/recreated per the rules
+  below.
+
   omac closes this by **executing from an approval snapshot**: at approval
   time the whole skill tree is frozen into a host-only directory the sandbox
   cannot write (`~/.config/omac/skills/<name>/<hash>`), and the sidecar is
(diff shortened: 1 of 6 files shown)


nhuelstng (approved):

BenediktBurger:
Thanks for your approval, @nhuelstng , however, I do not have permission to merge, that someone else has to do.

PR #251, opened 2026-08-20
fix(opencode): grant opentui grammar cache

### LINKED ISSUE #244: opencode harness: markdown renders as raw text outside $HOME (opentui tree-sitter grammars not granted)
### Duplicate check

### Context / Summary

The opencode harness renders markdown correctly when cwd is $HOME, but inline
styling breaks in any other directory. Root-caused to a missing filesystem grant
for opentui's tree-sitter grammar cache. Discovered while running omac locally on
macOS; no prior issue/PR.

### Problem / What

With the opencode harness, when cwd ≠ $HOME:
- Broken (rendered as raw text): bold, italic, bold-italic, strikethrough,
  inline code, code blocks, links, task lists.
- Still fine: blockquotes, tables.
- Works correctly when cwd = $HOME.

Root cause:
OpenCode's TUI uses opentui (github.com/sst/opentui), which loads tree-sitter
grammar .wasm files cached under ~/.local/share/opentui/tree-sitter/. The opencode
harness grants the sibling ~/.local/share/opencode but NOT ~/.local/share/opentui.
- cwd = $HOME: the whole-home workdir grant covers ~/.local/share/opentui → works.
- cwd elsewhere: ~/.local/share/opentui is denied → grammars can't load →
  markdown silently falls back to raw text.

The failure is silent: opentui/opencode swallows the read error, so nothing
appears in `omac diagnose` (0/4 blocked) or the network audit log, which is why
it was hard to diagnose. A visible symptom is that markdown briefly flashes
styled, then reverts to raw text (first paint, then the grammar-dependent
re-render fails).

Grant list to change: path/to/file.go:NN (the entry granting ~/.local/share/opencode).

### Suggested fix / Ask

Add ~/.local/share/opentui (read/write) to the opencode harness's default
filesystem grants, alongside ~/.local/share/opencode.

Read/write (not read-only) because opentui *downloads* and caches tree-sitter
grammar .wasm files on first use. On a fresh machine the directory is empty or
absent, so opentui must create and populate it; a read-only grant would
reintroduce the same silent failure on first run.

- [ ] Grant ~/.local/share/opentui (read/write) in the opencode harness profile (path/to/file.go:NN)
- [ ] Update docs/HARNESSES.md if it enumerates harness grants
- [ ] Verify markdown renders correctly from a non-$HOME directory (populated cache)
- [ ] Verify first-run works from a non-$HOME dir with an empty/absent ~/.local/share/opentui

Note: first-run download also requires network egress to the grammar host. If
that host isn't already reachable under the default filtered policy, it should be
allow-listed (or confirmed to work via the network prompt) so first-run isn't
silently broken.

### Non-goals

- Not changing opentui/opencode itself; the fix is purely an omac harness grant.
- Not touching non-opencode harnesses.
- Not widening the general filesystem policy beyond this one path.
- Not addressing the silent error-swallowing inside opentui (upstream concern).

### Evidence / Environment

Environment:
- OS: macOS
- omac version: 0.7.1
- OpenCode: 1.18.18
- Harness: opencode

Repro:
1. `omac start opencode` from a directory that is NOT $HOME
2. Ask the agent to reply with **bold**, `inline code`, and a fenced code block
3. Observe raw text (briefly flashes styled, then reverts)
4. Repeat from $HOME → renders correctly

Ruled out during diagnosis:
- Network: `omac diagnose` → 0/4 connection(s) blocked; audit log clean
- Color/env: TERM=xterm-256color, COLORTERM=truecolor, CLICOLOR_FORCE=1,
  FORCE_COLOR=1 all set inside sandbox → no effect
- Terminal/transport: raw `printf '\e[1mBOLD\e[0m \e[9mSTRIKE\e[0m'` renders
  correctly inside the sandbox
- `-no-sandbox` renders correctly → cause is the sandbox filesystem scope

Confirmed fix (local):
Added ~/.local/share/opentui (read) to the sandbox profile → markdown renders
correctly from a non-$HOME directory.

NoRiceToday:
**Issue:** Closes #244

## What
- Restores OpenCode markdown rendering outside `$HOME`, including first-use grammar cache setup under default and custom XDG data roots.

## Why
- OpenTUI caches tree-sitter grammars outside OpenCode’s existing sandbox grants, so rendering silently fell back to raw text.

## How
- Resolve OpenTUI’s cache from `$XDG_DATA_HOME`, falling back to `~/.local/share/opentui`.
- Create the cache before `start` and `serve` grant resolution, then grant it read/write only for OpenCode.
- Enforce that pre-created directories are also sandbox-granted and document the new grant.

## Verification
- `go test -buildvcs=false ./... -skip ^TestIntegration -count=1` — pass
- `go build -buildvcs=false ./...` — pass
- `go vet ./internal/config ./internal/cli ./internal/sandboxprofile` — pass
- `go test -buildvcs=false -tags=e2e ./internal/e2e -run ^$ -count=1` — pass (compile check)
- `git diff --check` — pass

## Follow-up
- Full macOS kernel-sandbox integration is covered by CI; nested local Seatbelt application is denied by the active sandbox.

🤖 Generated with OpenCode (GPT-5.6 Sol).

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -70,7 +70,7 @@ cannot access.
 | Path or variable | Access | Why |
 |---|---|---|
 | `<workdir>` | read + write | Your project files |
-| Harness config dirs (e.g. `~/.claude`, `~/.local/share/opencode`) | read + write | The harness stores its state and credentials here |
+| Harness config dirs (e.g. `~/.claude`, `~/.local/share/opencode`, `~/.local/share/opentui`) | read + write | The harness stores its state and credentials here; omac pre-creates declared first-use dirs (e.g. OpenCode's `opentui` tree-sitter grammar cache) before sandbox grant resolution |
 | omac-managed tool cache (isolated from `~/.cache`; see [Cache](./advanced/cache.md)) | read + write | Build artifacts and downloaded packages; isolated from your host caches |
 | Language toolchain binaries (`~/.cargo/bin`, `~/go/bin`, `~/.nvm`, `~/.bun/bin`, `~/.rustup`) | read-only | So installed compilers and build tools can run |
 | Shared skills dirs (`~/.config/agents/skills`, `~/.agents/skills`) | read-only | So the agent can read skill descriptions (`SKILL.md`) |
--- a/internal/cli/continue_resume_test.go
+++ b/internal/cli/continue_resume_test.go
@@ -389,6 +389,7 @@ func launchCacheCaptureForHarness(t *testing.T, harnessName string, noSandbox, e
 	home := t.TempDir()
 	t.Setenv("HOME", home)
 	t.Setenv("XDG_CONFIG_HOME", filepath.Join(home, ".config"))
+	t.Setenv("XDG_DATA_HOME", "")
 	for key := range toolcache.Environment("ignored", toolcache.ModePersistent) {
 		value := "host-" + key
 		if harnessName == "opencode" {
--- a/internal/cli/continue_resume_test.go
+++ b/internal/cli/continue_resume_test.go
@@ -330,6 +330,21 @@ func TestLaunchCacheOmitsVerboseOutputWithoutVerbose(t *testing.T) {
 	}
 }
 
+func TestLaunchOpenCodeCreatesAndGrantsRuntimeDirs(t *testing.T) {
+	capture := launchCacheCaptureForHarness(t, "opencode", false, false, false)
+	opentuiDir := filepath.Join(capture.home, ".local", "share", "opentui")
+	if info, err := os.Stat(opentuiDir); err != nil || !info.IsDir() {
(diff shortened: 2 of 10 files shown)


nhuelstng:
**Should fix** : test escapes the temp home when XDG_DATA_HOME is set. TestLaunchOpenCodeCreatesAndGrantsRuntimeDirs (internal/cli/continue_resume_test.go:266) doesn't neutralize an ambient XDG_DATA_HOME. I verified: with XDG_DATA_HOME=/some/dir the test fails and prepareSandboxDirs creates /some/dir/opentui on the developer's real machine. The sibling tests in serve_test.go/harness_test.go do t.Setenv("XDG_DATA_HOME", ""); launchCacheCaptureForHarness needs the same (or point it under home).

nhuelstng (approved):

Issue #241, opened 2026-08-17
Skainet available model update not working


leimer:
### Duplicate check

- [x] I have searched the existing issues

### Context / Summary

I am trying to update my models list available in Skainet to use them in opencode with omac.

### Problem / What

What I tried so far:

Removing the installation and caches of @tngtech/opencode-skainet
- rm -rf ~/.bun/install/cache/.registry-cache 2>/dev/null
- rm -rf ~/.bun/install/cache/@tngtech
- rm -rf ~/.local/share/opencode/plugin
- omac start -> no models synced

in omac config folder, check for outdated version
- npm view  @tngtech/opencode-skainet versions --json
- npm view  @tngtech/opencode-skainet version
- version is on latest 1.0.3

reinstall opencode plugin directly
- opencode plugin --global @tngtech/opencode-skainet@$(npm view @tngtech/opencode-skainet version) --force
- opencode auth login
- no success

### Suggested fix / Ask

What can I do to have the currently available models usable in my omac opencode harness?

### Evidence / Environment

╰─❯ omac doctor
omac 0.7.1
OS: macos
workdir: /Users/johannes/workspace/iswe/repos/invoice-processor
[ok] keychain backend: reachable
[ok] launcher config: (built-in defaults)
[ok] registry: 2 skill(s) registered (0 workdir, 2 global)
  [ok] gradlew              binary=yes missing_required_secrets=0
  [ok] skill-marketplace    binary=yes missing_required_secrets=0
Inner harnesses:
  [ok]   opencode     binary=opencode found
  [warn] claude-code  binary=claude not on $PATH
  [warn] codex        binary=codex not on $PATH
  [warn] copilot      binary=copilot not on $PATH
  [warn] pi           binary=pi not on $PATH
  [warn] codewhale    binary=codewhale not on $PATH
  [ok] built-in omac-write-a-skill (opencode): present
[ok] sandbox profile "builtin" uses the built-in sandbox
[ok] built-in sandbox: kernel backend available
[ok] network prompt: dialog backend available
sandbox profile lint (2 finding(s), advisory):
  [medium] filesystem.read: overlaps known secret path /Users/johannes/.gitconfig not in baseline (git config (may embed tokens)) (~/.gitconfig)
  [medium] network.allow_tcp_connect: direct outbound TCP to SSH/RDP port (22)

When a run fails, `omac diagnose` shows what the sandbox blocked and why.

Retr0MrWave:
[...]

Issue #155, opened 2026-07-21
Generic per-harness cache pre-warm before sandboxing (first-launch plugin/dep fetch fails through omac's proxy; opencode is the first concrete case)


nhuelstng:
## Problem (general)

Any harness that fetches something from the network on its **first launch** — a plugin, a provider package, model metadata, a runtime component — breaks the first time it runs under `omac start` in a new workdir, because:

1. **Authenticated proxy not traversed.** omac injects `HTTP_PROXY`/`HTTPS_PROXY=http://omac:<token>@127.0.0.1:<port>` and forces all egress through its filtering proxy. Installers/fetchers that don't send the proxy's basic-auth credentials fail **silently** — no `net DENY` line, because the request never becomes an allowed connection. `--allow-domain` on the registry does not help; it is not a filesystem grant issue either.
2. **Per-workdir cache.** The tool cache is `PreparePersistent(DomainWorkdir, …)` → `~/.cache/omac/sha256(\"v1:workdir:\"+canonical-workdir)`, and `omac start`/`serve` inject `XDG_CACHE_HOME=<scope>/xdg` (internal/toolcache/cache.go, internal/cli/start.go). So the cache is **per-workdir**: every new project folder starts cold and re-hits the failing fetch. A one-time global warm does not help.

The harness works fine **outside** omac (direct fetch), but fails on the first `omac start` in any workdir the user has not already warmed. Since we cannot know in advance which folder the user will start omac in, a manual/install-time pre-warm of one workdir is not a reliable fix.

## First concrete case: OpenCode `@tngtech/opencode-skainet`

OpenCode config declares `plugin: [\"@tngtech/opencode-skainet@latest\"]`; on startup it uses Bun's package installer to fetch the plugin, then loads it to register the `skainet` provider. Inside the sandbox the install fails (per above), so the provider is never registered:

```
ProviderModelNotFoundError: Model not found: skainet/zai-org/GLM-5.2
```

(surfaced as a generic `UnknownError / Unexpected server error`).

### Reproduce

```
XDG_CACHE_HOME=$(mktemp -d)/xdg \
  omac sandbox run --profile tng-default --allow-domain chat.model.tngtech.com \
    --allow-env XDG_CACHE_HOME --write <config> --write <cache> \
    -- opencode run \"hello\"
# -> ProviderModelNotFoundError: Model not found: skainet/...
```

[...]

PR #259, opened 2026-08-24
feat(sandbox,cli): project a credential-free registry config into the sandbox

### LINKED ISSUE #150: feat(sandbox): support & detect private-registry auth (protected ~/.npmrc)
## What

Give harnesses a supported, discoverable way to authenticate against a **private package registry** from inside the sandbox, and detect when a private-registry auth token is present but unreachable because the file that holds it is protected.

Today `~/.npmrc` is a **protected path** (`internal/sandboxprofile/baseline.go:51`, `protectedCommon()`), classified as an "npm token" secret (`internal/profileaudit/check.go:141`). It is denied by default even under a broader home-dir grant. It is *not* hard-blocked — it is overridable via `filesystem.override_deny` (`EffectiveProtectedPaths`, `baseline.go:188`) — but nothing surfaces that escape hatch, and the failure it produces is silent.

## Why

This is the **filesystem/credential** counterpart to #148 (which explains blocked registry *fetches* on the network axis). The two failures stack for the private-registry case:

1. #148 covers the network: the registry host is denied or unreachable (VPN down).
2. **This issue** covers auth: even when the host *is* reachable and allowed, `npm`/`pip`/`cargo` must read the auth token from `~/.npmrc` (or `~/.pypirc`, cargo credentials, …), and that read is denied by the protected-path policy. The install fails `401`/`403` **at the registry**, not at the proxy.

Related to #143 (harness fetching a config-declared plugin into a cold cache): a private-registry plugin can clear the network gate and still fail auth, with no legible cause.

The protection itself is correct — the token is a secret and must not leak to arbitrary sandboxed code. The gap is that there is (a) no documented, scoped way to opt a real private-registry workflow in, and (b) no signal that this is what went wrong.

## Detection — can omac detect it?

**Not at runtime via the proxy.** For HTTPS the proxy splices a raw TCP tunnel and TLS bytes pass through unread (`internal/netproxy/server.go:266`, `handleConnect`). It is not a MITM, so it cannot observe the registry's `401`/`403` auth response or body. The `#148` hints fire on the proxy's *own* deny (`403 denyBody`) or dial failure (`502 upstreamErrorBody`) — a registry-issued auth failure is invisible to it. So we cannot bolt private-registry-auth detection onto the #148 network path.

**Yes, proactively at launch — this is the tractable path.** Mirror #148's `isPackageRegistry(host)` host detector on the *config file* instead of the wire:

- At sandbox setup, if `~/.npmrc` exists, is **not** listed in `filesystem.override_deny`, and contains a registry-scoped auth entry (`_authToken`, `//<host>/:_authToken`, `_auth`, `_password`) pointing at a **non-public** registry (reuse the private-registry heuristic from #148's `isPackageRegistry`), emit an actionable note:
  > `~/.npmrc` holds a private-registry token but is protected by the sandbox; private-package installs will fail auth. To allow it, add `~/.npmrc` to `filesystem.override_deny`, or supply the token via `NPM_TOKEN`/`NODE_AUTH_TOKEN` through the env allowlist.
- Same shape generalizes to `~/.pypirc` (`check.go:147`) and cargo credentials.

This is a static, read-side heuristic used only to enrich a likely-failing setup, so a false positive is harmless — same design principle #148 applied to its host matcher.

## Proposed remedies to document/support

- **Escape hatch (exists, undocumented):** `filesystem.override_deny: ["~/.npmrc"]` punches the specific hole while keeping every other credential protected.
- **Env-based token (preferred, no file grant):** forward `NPM_TOKEN`/`NODE_AUTH_TOKEN` (and pip/cargo equivalents) via the env allowlist and let the harness synthesize a minimal `.npmrc` in the workdir — no home-dir secret exposure. Aligns with the env-forwarding pattern already used for provider creds.
- Document both in the sandbox profile reference alongside the protected-path list.

## Scope

- [ ] Launch-time detector + hint for a protected-but-populated `~/.npmrc` (npm first; pypirc/cargo as follow-ups).
- [ ] Reuse / share #148's private-registry heuristic so the file-side and host-side detectors agree on what "private registry" means.
- [ ] Docs: private-registry auth section — `override_deny` vs env-token, with the security tradeoff spelled out.

## Non-goals

- Auto-granting `~/.npmrc` (would defeat the secret protection).
- Runtime MITM of registry auth responses (proxy is a passthrough tunnel by design).

Related: #148 (network-axis registry hints), #143 (cold-cache plugin fetch).

Retr0MrWave:
**Issue:** Closes #150 · Refs #241

## What

- New opt-in profile knob `filesystem.registry_config: ["npm"]`: omac projects a **credential-free** copy of `~/.npmrc` into the sandbox so scoped packages resolve against their private registry.
- `omac doctor` now reports a private registry mapping the sandbox cannot see — the failure that previously had no diagnostic surface anywhere.
- Unset means today's behaviour: the file stays fully masked.

## Why

`~/.npmrc` is a baseline protected path (`internal/sandboxprofile/baseline.go:51`) because it commonly holds `_authToken`. But the same file carries the `@scope:registry=` **mapping**, which is load-bearing configuration, not a secret. With it masked, npm resolves a scoped package against the public registry and gets a 404 that reads like *"no such package"* — no denial event, nothing in `omac diagnose`, and (per #241) host-side cache purges and `--force` reinstalls have no effect because they touch paths the sandbox never reads.

The blunt existing workaround, `override_deny: ["~/.npmrc"]`, works but grants the whole file including the token — exactly what #150 set out to avoid.

## How

- Profile knob mirrors the existing `network.proxy_injection` mechanism: accepted-value list next to `ProxyInjectionTools()`, rejected-if-unknown in `Profile.Validate()`.
- New `internal/registryconf` does the scrub. Only registry-mapping keys survive; a kept value must resolve to a credential-free `http(s)` URL. npm's own value syntax is honored first (quotes stripped, `${VAR}` expanded) so a mapping npm would act on is not lost.
- `internal/sandboxrun/registryconf.go` writes the copy per launch, appends **only that file** to `Grants.ReadPaths`, and injects `NPM_CONFIG_USERCONFIG` through the same `injected` env map as `ProxyInjectionEnv`.
- The doctor check reads the filesystem, so it deliberately does **not** live in `profileaudit.Check` (documented I/O-free) — it sits alongside the other doctor checks.

Two things are refused rather than projected, both reported at launch and by doctor:

- A URL with a query string or fragment — that is where an API key usually hides (`?apiKey=…`) and omac cannot tell a secret parameter from a load-bearing one.
- The **global** `registry` key when the file holds a credential for that host: redirecting all resolution to a registry omac cannot authenticate to would break even the public installs that work today. A scoped mapping to such a host is kept (that scope was already failing) with a 401/403 warning.

Note: the file-side detector uses "host is not npm's default" rather than netproxy's `isPackageRegistry` heuristic, which returns **false** for exactly the corporate shape this exists for (`tng-artifacts.int.tngtech.com` has no `registry`/`npm` label) — verified before relying on it.

## Verification

`go build ./...`, `go vet ./internal/...` clean. `go test ./...` passes except `TestIntegrationWorktreeKnownLimitations` and `TestIntegrationWorkflowInterpretersRunnable`, which fail identically on clean `main` on this host (local toolchain paths absent from the default profile).

End-to-end against #241's actual case, Linux/bwrap, **one binary, cold cache, no `override_deny`**:

```
knob off -> 0 skainet models   (only ALLOW registry.npmjs.org — the 404 path)
knob on  -> 14 skainet models, installs @tngtech/opencode-skainet 1.0.3
            ALLOW tng-artifacts.int.tngtech.com:443, ALLOW chat.model.tngtech.com:443
```

Inside the sandbox with the knob on:

```
cat ~/.npmrc               -> No such file or directory   (host file still masked)
cat $NPM_CONFIG_USERCONFIG -> @tngtech:registry=https://tng-artifacts.int.tngtech.com
```

`omac doctor` on a profile without the knob names the cause unprompted:

```
[warn] registry config: /home/…/.npmrc maps a scope to tng-artifacts.int.tngtech.com,
       but the sandbox cannot read it
       Scoped installs will fail with a 404 against the public registry. Fix:
       add filesystem.registry_config: ["npm"] to the sandbox profile.
```

The second commit is review fixes; each finding has a regression test (query-string leak, npm value syntax, unauthenticated global registry, non-fatal read failure, doctor reporting both projection and override).

## Follow-up

- #241 stays open as the user-facing anchor until the reporter confirms. Their steps are the knob plus `omac cache clear` (the omac cache scope pins `@latest` at first resolution, so a stale scope survives the fix).
- #155 should be re-scoped against this: its opencode case is this bug, and a pre-warm would only mask it by warming outside the sandbox. Its per-workdir premise is also stale post-#144.
- Not addressed here: the plugin's model cache at `$XDG_STATE_HOME/opencode-skainet/`, which omac does not grant (`internal/config/harness.go:253`).


internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+		return "", "", false, "URL carries a query string, which commonly holds an API key omac must not copy into the sandbox"
+	}
+	if u.Fragment != "" {
+		return "", "", false, "URL carries a fragment"
+	}
+	if u.User != nil {
+		u.User = nil
+		stripped = true
+	}
+	return u.String(), u.Hostname(), stripped, ""
```
NoRiceToday: [MUST FIX] bug, security, missing-test

`registryURL` returns `u.Hostname()` (port stripped), but `credentialHost` (line 401) keeps `host:port`, so a port-scoped credential never correlates with its mapping:

```
registry=https://npm.acme.test:8443
//npm.acme.test:8443/:_authToken=T
```

yields `kept=[registry], needsAuth=[]` — the global registry to a credentialed host is projected, and the scoped mapping to that host silently loses its NeedsAuth warning. This is the regression eae1299 claims to close; its regression test only covers the portless shape.

```suggestion
	return u.String(), u.Host, stripped, ""
```

Correlating on `u.Host` is contained — the host feeds only `keptHosts` correlation and rejection message text. Add the port pair (global refused, scoped flagged) to `TestScrubNPMRCRefusesUnauthenticatedGlobalRegistry`.

internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+// good mapping (`@acme:registry="https://npm.acme.test"`) would look
+// unparseable and be silently skipped.
+func npmValue(value string) string {
+	v := strings.TrimSpace(value)
+	if len(v) >= 2 {
+		if (v[0] == '"' && v[len(v)-1] == '"') || (v[0] == '\'' && v[len(v)-1] == '\'') {
+			v = v[1 : len(v)-1]
+		}
+	}
+	return os.ExpandEnv(v)
```
NoRiceToday: [SHOULD FIX] security, bug, docs-drift

`os.ExpandEnv` expands from the full host env, and an expanded secret survives into the sandbox-readable projection in a URL path position: `@acme:registry=https://npm.acme.test/api/${SECRET}/npm/` with `SECRET=LIVESECRET` projects `…/api/LIVESECRET/npm/`. Query/fragment are refused for exactly this reason; path is not, so the package invariant "no credential can survive by construction" does not hold.

Refusing expanded values outright would regress the corporate `${ART_HOST}` shape the expansion exists for. Suggested: when expansion actually modified the value, emit a launch/doctor warning of the same class as `Rejected` ("mapping was expanded from `${…}`; verify it carries no secret"), and document the residual path-position risk next to the query/fragment refusal.

internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+	NeedsAuth []string
+}
+
+// scopedRegistryKey matches npm's scoped-registry form, e.g.
+// "@acme:registry" or "@acme/sub:registry".
+var scopedRegistryKey = regexp.MustCompile(`^@[^:\s]+:registry$`)
+
+// credentialKey matches npmrc keys that carry authentication material,
+// including the per-registry form "//host/path/:_authToken".
+var credentialKey = regexp.MustCompile(`(?i)(_auth|_authtoken|_password|username|email|^//)`)
```
NoRiceToday: [NICE TO HAVE] bug, docs-accuracy

The `^//` alternative counts `//npm.acme.test/:always-auth=true` as credential material (`DroppedCredentials++`), so doctor prints "including the auth token it holds" for a file holding only a boolean flag. Restrict the classification to `_auth|_authtoken|_password` variants; treat other `//host/:` keys as dropped-but-not-credential.

internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+			v = v[1 : len(v)-1]
+		}
+	}
+	return os.ExpandEnv(v)
+}
+
+// credentialHost extracts the registry host from a per-registry auth key
+// such as "//npm.acme.test/:_authToken" or "//npm.acme.test/api/:_password".
+// Returns "" for keys that are not host-scoped (e.g. a bare "_auth").
+func credentialHost(key string) string {
```
NoRiceToday, line 503: [NICE TO HAVE] decision-conflict

Host-less credentials (`_auth=…`, `_password=…`) return `""` here and never correlate with the global `registry` mapping — `registry=https://npm.acme.test` + `_auth=BASE64SECRET` is kept with no NeedsAuth flag. That contradicts the stated refusal rationale ("the file held a credential for that host"), and eae1299's message claims credential correlation that only holds for `//host/`-scoped keys. Pin the chosen semantics in a comment + test so this doesn't get re-filed.

internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+	var res ScrubResult
+	var lines []string
+	// keptHosts maps a kept mapping key to its registry host, so the
+	// credential correlation below can run after the whole file is read
+	// (auth lines may appear before or after the mapping they apply to).
+	keptHosts := map[string]string{}
+	authHosts := map[string]bool{}
+
+	for _, raw := range strings.Split(string(src), "\n") {
+		line := strings.TrimSpace(raw)
```
NoRiceToday, line 356: [NICE TO HAVE] bug

A UTF-8-BOM npmrc (plausible for Windows-authored files) has its leading `@scope:registry=…` counted as Dropped rather than Rejected: `hosts=[]`, `Rejected=[]` → `InspectNPM` returns nil → neither launch nor doctor says anything — the silent 404 this feature exists to prevent. Trim a leading BOM once, or report leading garbage as Rejected.

internal/registryconf/registryconf.go
```diff
@@ -0,0 +1,446 @@
...
+// would act on is not silently lost. Anything still unusable — or carrying
+// a secret outside the userinfo, e.g. ?apiKey= — is recorded in Rejected
+// rather than dropped quietly.
+func ScrubNPMRC(src []byte) ScrubResult {
+	var res ScrubResult
+	var lines []string
+	// keptHosts maps a kept mapping key to its registry host, so the
+	// credential correlation below can run after the whole file is read
+	// (auth lines may appear before or after the mapping they apply to).
+	keptHosts := map[string]string{}
```
NoRiceToday: [NICE TO HAVE] maintainability

Duplicate mapping keys project twice, and the correlation loop pairs `res.KeptKeys[i]` with `lines[i]` by index across a re-filtering pass — a future `continue` between the two append sites (lines 341-342 vs 368-369) desyncs them silently. Carrying `mapping{key, host, line}` in one slice (with explicit last-wins) removes both hazards.

internal/sandboxrun/registryconf.go
```diff
@@ -0,0 +1,88 @@
...
+			fmt.Fprintf(stderr, "omac sandbox: WARNING: registry_config %s: %s\n", p.Ecosystem, p.Warning)
+		}
+		if !p.Projected() {
+			continue
+		}
+
+		// Grant exactly the projected file, read-only. The host file is
+		// untouched and stays protected.
+		grants.ReadPaths = append(grants.ReadPaths, p.Path)
+		injected[p.EnvVar] = p.Path
```
NoRiceToday, line 74: [NICE TO HAVE] backwards-compat

A profile that both allowlists `NPM_CONFIG_USERCONFIG` and opts into the projection silently gets the projection instead — `injected` wins, and the user's other settings in their config are dropped. Emit a one-line warning when `injected` already contains `NPM_CONFIG_USERCONFIG`.

internal/cli/doctor.go
```diff
@@ -199,6 +206,85 @@ func doctorProfileLint(env *Env, profileRef string) {
...
+//
+// Advisory only — it never affects doctor's exit code.
+func doctorRegistryConfig(env *Env, profileRef string) {
+	profile, _, err := sandboxprofile.Resolve(profileRef)
+	if err != nil {
+		return // profile problems are already reported by the sandbox section
+	}
+	enabled := slices.Contains(profile.Filesystem.RegistryConfig, sandboxprofile.RegistryConfigNPM)
+	src, err := registryconf.NPMUserConfig()
+	if err != nil {
```
NoRiceToday, line 276: [NICE TO HAVE] bug, consistency

The launch path turns a non-ENOENT read failure on `~/.npmrc` (e.g. EACCES after a root-owned `sudo npm config set`) into a projection warning (registryconf.go:143-145), but doctor swallows the identical error and prints nothing. Report it as `[warn]` or reuse the projector's Warning path.

internal/sandboxprofile/profile.go
```diff
@@ -114,6 +114,22 @@ type Filesystem struct {
...
+	// This exists because a scope→registry mapping is load-bearing
+	// *configuration* that happens to live in the same file as a
+	// credential. Without it, a scoped package resolves against the
+	// public registry and 404s (see #150, #241). Values:
+	//   - "npm": derive from ~/.npmrc, inject NPM_CONFIG_USERCONFIG.
+	//
+	// Unset means the historical behavior: the file stays fully masked.
+	// The blunt alternative — override_deny on ~/.npmrc — grants the
+	// whole file including any auth token; this does not.
+	RegistryConfig []string `json:"registry_config,omitempty"`
```
NoRiceToday, line 132: [NICE TO HAVE] docs-drift, backwards-compat

Adding this field means an older omac binary hard-rejects any shared/base profile that gains `registry_config` with a bare `json: unknown field "registry_config"` (strict decode at profile.go:301) — no hint that a newer omac is needed, so upgrading one machine breaks every other machine's profile. Consider a known-field hint in the parse-error path, and note the forward-incompatibility in CONFIGURATION.md.

NoRiceToday:
**Adversarial review — verdict: fix-then-ship, risk: medium.** Findings 1-9 inline; P1/P2 confirmed empirically against the real scrubber.

**Test gaps (nice to have):** the P1/P2 both slipped through because there is no port-scoped credential case and no post-condition pinning that env-expanded values are scrutinized after expansion. Also missing: duplicate-key, BOM, and CRLF cases (CRLF currently correct but untested), a pre-set `NPM_CONFIG_USERCONFIG` case, a symlinked `~/.npmrc` case, and Windows `USERPROFILE` staging (tests use `t.Setenv("HOME")`, which `os.UserHomeDir` does not read on Windows).

internal/registryconf/registryconf.go
Retr0MrWave: Fixed, and confirmed the shape you gave was projected before the fix.

Took `u.Host` as suggested, but routed both sides through one `normalizeHost`
helper rather than only changing the return, because the same comparison had a
second break: `credentialHost` lowercased and `u.Hostname()` did not, so
`registry=https://NPM.Acme.Test` + `//npm.acme.test/:_authToken=T` also failed
to correlate. Normalization is lowercase + strip an explicit default port
(npm's nerfDart does the same), so `https://host:443` now correlates with
`//host/:_authToken` too.

Tests: `TestCorrelationMatchesPortScopedCredential` (your exact pair, global
refused *and* the scoped variant flagged), `TestCorrelationIsCaseInsensitive`,
`TestCorrelationNormalizesDefaultPort`, and `TestCorrelationIgnoresUnrelatedHost`
for the flip side.

internal/registryconf/registryconf.go
Retr0MrWave: Fixed, and I went stricter than your suggestion: a placeholder outside the URL's
authority is **refused**, not warned about.

Reasoning: a warning leaves the package's stated invariant conditional on
someone reading it, and we already refuse a query string for exactly this
"cannot tell a secret from a load-bearing value" reason — so warning here would
be inconsistent with the neighbouring rule. Your concern about regressing
`${ART_HOST}` is addressed by keeping expansion in the authority:

```
@acme:registry=https://${ART_HOST}/api/npm/npm/   -> projected
@acme:registry=https://host:${PORT}/npm/          -> projected
@acme:registry=https://${U}:${PW}@host/npm/       -> expands, then userinfo stripped
@acme:registry=https://host/api/${SECRET}/npm/    -> refused, names SECRET
```

Mechanism: split the raw template at the authority boundary (find `"://"`, then
the first `/`, `?` or `#`) — structural, because an unexpanded template usually
does not parse as a URL. The authority goes through `os.ExpandEnv`; the
remainder goes through `os.Expand` with a mapper that only records that a
placeholder was consumed. So the refusal decision is made by `os.Expand`'s own
tokenizer (`${NAME}`, `$NAME`, `$$` cannot drift from the expansion path), and
no expanded *value* is ever matched against the URL — no coincidental-substring
false positives. `TestExpansionAllowedInAuthorityOnly` covers each position,
including a literal `$` in the path that must **not** trigger a refusal.

Docs and the package doc updated: the invariant now reads as "userinfo removed;
query, fragment and non-authority interpolation refused", which is what the code
actually enforces.

internal/registryconf/registryconf.go
Retr0MrWave: Fixed. Classification now matches the key's **leaf** (after the last `:`)
against `_auth|_authtoken|_password`, so `//host/:always-auth=true` is dropped
like any other non-mapping key but no longer counted, and no longer marks the
host as needing auth. `username`/`email` are also excluded — they are dropped,
but they are not tokens and counting them overstated what a projection
protects. `TestHostScopedNonCredentialIsNotCountedAsCredential`.

internal/registryconf/registryconf.go
Retr0MrWave: Fixed by choosing the semantics rather than only documenting them: npm applies
legacy host-less `_auth`/`_password` to the **default** registry, so they now
bear on the global `registry` mapping (refused, same as the `//host/`-scoped
case). A scoped mapping is unaffected — a host-less credential says nothing
about a scope's own registry.

Pinned by comment and by `TestHostlessCredentialAppliesToGlobalRegistry`, which
asserts both halves so the alternative reading cannot be re-filed silently.

internal/registryconf/registryconf.go
Retr0MrWave: Fixed: one leading BOM is trimmed at the top of `ScrubNPMRC`. Good catch on the
consequence — it was not just a mis-count, it landed in `Dropped` where
`InspectNPM` returns nil, so neither the launch path nor doctor said anything,
which is precisely the silent 404 this feature exists to prevent.
`TestScrubHandlesBOM`. (`TestScrubHandlesCRLF` pins the CRLF behaviour you
noted as correct-but-untested, including that the CR does not defeat
correlation.)

internal/registryconf/registryconf.go
Retr0MrWave: Fixed as suggested: one `[]mapping{key, host, line}` replaces the parallel
slices, so nothing can desync, and duplicates are resolved explicitly with
last-wins (npm's ini semantics) instead of being projected twice.
`TestDuplicateKeyLastWins`.

internal/sandboxrun/registryconf.go
Retr0MrWave: Fixed: warns when the variable is already set to a different path, naming the
path whose settings the sandbox will not see. The projection still takes
precedence — the warning explains rather than defers, since silently honouring
the user's file would reintroduce the unmasked-config problem. Tests for both
the warning and the quiet ordinary path.

internal/cli/doctor.go
Retr0MrWave: Fixed: reported as `[warn]` with the consequence spelled out, matching the
launch path. Worth flagging that doctor is the *only* check that runs before a
launch, so this silence was the more costly of the two.
`TestDoctorRegistryConfigReportsUnreadableConfig`.

internal/sandboxprofile/profile.go
Retr0MrWave: Both parts done. `CONFIGURATION.md` gains a note that profiles are parsed
strictly and a profile using `registry_config` is rejected by an older omac, so
shared/checked-in profiles want the upgrade first. And the unknown-field branch
of `Parse` now suggests which direction to look ("if this profile was written
for a newer omac, upgrade omac; otherwise remove the field") while other parse
errors keep the plain message. Strictness is unchanged — a typo must still fail
loudly. `TestUnknownFieldErrorHintsAtVersionSkew` asserts both the hint and that
it does not leak onto unrelated parse errors.

Note this only helps from this version forward; already-released binaries will
still emit the bare message. That is why the doc note matters more than the code
change here.

Retr0MrWave:
Thanks — this was a genuinely useful review. All nine findings are fixed in
`c804778`; each has a regression test. Two notes on where I went further or
narrower than suggested:

- **P1 had a second facet.** Beyond the port mismatch, `credentialHost`
  lowercased its host while `u.Hostname()` did not, so a mixed-case mapping
  (`registry=https://NPM.Acme.Test`) also escaped correlation. Both sides now
  normalize through one helper: lowercase plus strip an explicit default port,
  matching npm's own nerfDart keying. Default ports are stripped regardless of
  scheme *deliberately* — a credential key carries no scheme, so a scheme-aware
  rule could only apply to one side, and any asymmetry there risks
  **under**-correlating, which silently reopens the hole. The one inaccuracy
  that accepts (`http://host:443` matching a `//host/` credential) fails safe:
  it refuses a mapping and says why.
- **P2 is refused rather than warned.** Rationale below.

On the test-gap list: added port-scoped, mixed-case, default-port and
unrelated-host correlation; expansion in host/port/userinfo/path positions plus
a literal `$`; boolean host-scoped key; host-less credential for global vs
scoped; BOM; CRLF; duplicate keys; symlinked `~/.npmrc`; pre-set
`NPM_CONFIG_USERCONFIG`; doctor read failure; and the unknown-field hint.

**Windows staging: deliberately not added.** You are right that
`os.UserHomeDir` ignores `HOME` there, but `.goreleaser.yaml` builds `linux`
and `darwin` only, so there is no Windows target to protect and Windows test
scaffolding would be dead weight. The helper now sets `USERPROFILE` alongside
`HOME` — one line, so the helper does not misrepresent what it stages if that
ever changes.

NoRiceToday (approved):
LGTM now, thanks for addressing the feedback :) Just take care of the conflicts then you can merge

PR #260, opened 2026-08-24
feat(netproxy,diagnose): record network prompts nobody answered in time

### LINKED ISSUE #257: Network prompt latency silently defeats short client timeouts, and an abandoned prompt leaves no record
### Duplicate check

### Context / Summary

Found while reproducing #241 (opencode's skainet plugin ships no models under
omac). Independent of that root cause (#150), the same runs exposed a generic
failure mode: a tool with a short HTTP timeout behind an interactive network
prompt gives up before the human answers, and **nothing anywhere records that
this happened**. Refs #241, #155, #148.

### Problem / What

Two coupled defects.

**1. Prompt latency exceeds ordinary client timeouts.** In filtered mode the
proxy holds the connection until a verdict. Measured on this box, answering a
GUI prompt took ~5 s (`intent-signal … models.dev` 09:58:27 → `ALLOW
models.dev:443 (prompt:allow)` 09:58:32) — that is a fast answer with the dialog
already in focus. The opencode skainet plugin aborts its model-discovery fetch
after **3 s** (`AbortSignal.timeout(getModelsTimeout())`, default
`DEFAULT_MODELS_TIMEOUT_MS = 3_000`). So the tool has already given up by the
time the user clicks Allow, and the connection then succeeds for a request
nobody is waiting on. The user's own click is what makes the run fail.

This is not an edge case: the compiled-in default profile
(`internal/sandboxprofile/resolve.go:23`) carries **no `allow_domain` at all**,
so on a fresh install every host prompts on first contact.

**2. An abandoned prompt is invisible.** In the run where the client gave up,
the audit log has the `intent-signal` line and then **no `net.decision` event
at all** — the verdict never lands, so nothing is recorded as allowed, denied,
or timed out. Consequences:

- `omac diagnose` on the failing run reports
  `last run exited 0 · 0/1 connection(s) blocked` and
  "Nothing was blocked by the network policy" — for a run whose plugin install
  silently failed.
- `decisionsFromEvents` (`internal/cli/diagnose.go:173`) only maps
  `net.decision` events, so there is no event type for this case to map.
- `omac sandbox run` does not write the central audit log at all, so
  reproductions driven that way leave no trail either.

Today's audit log after four repro runs contains exactly two `net.decision`
events, both `registry.npmjs.org allow=true source=prompt`.

The existing registry hints (#148, `internal/netproxy/registryhint.go`) can't
help: they fire on the proxy's *own* deny or dial failure. Here the fetch was
neither denied nor failed — it was abandoned client-side.

### Suggested fix / Ask

- [ ] Emit an audit event when a prompt is raised and the client disconnects
      before a verdict (`net.prompt_abandoned` or similar: host, port,
      time-to-abandon, whether a verdict later arrived).
- [ ] Surface it in `omac diagnose`: "a prompt for `<host>` was raised but the
      requesting tool gave up after `<N>`s — its own timeout is shorter than
      your answer time. Pre-allow the host with `network.allow_domain`."
- [ ] Stop `omac diagnose` reporting "nothing was blocked" when abandoned
      prompts exist for the analyzed run.
- [ ] Consider making `omac sandbox run` write the same audit trail as
      `start`/`serve`, so reproductions are diagnosable.
- [ ] Docs: note that short-timeout clients need their hosts pre-allowlisted,
      and that `on_unavailable: deny` turns this into a deterministic silent
      failure under `omac serve` / non-TTY.

### Non-goals

- Changing the plugin's 3 s timeout (upstream, not ours).
- Shipping TNG-specific hosts in the OSS default profile.
- Auto-allowing hosts to dodge the race.

### Evidence / Environment

omac built from `main` (26ff8e3), Linux/bwrap, opencode 1.17.16.

Abandoned prompt — control run, `models.dev` prompt raised, no verdict ever:

    2026-08-24 09:58:12 INFO ask intent-signal: network prompt host=models.dev port=443 intent=missing
    (no net.decision for models.dev; run exits 0)

Answered prompt — 5 s human latency vs the plugin's 3 s budget:

    2026-08-24 09:58:27 INFO ask intent-signal: network prompt host=models.dev port=443 intent=missing
    2026-08-24 09:58:32 INFO net ALLOW models.dev:443 (prompt:allow)

Audit log after the repro runs:

    net.decision: 2  (both registry.npmjs.org, allow=true, source=prompt)

`omac diagnose` on the failing run:

    last run exited 0 · filtered mode · 0/1 connection(s) blocked
    • Nothing was blocked by the network policy

Retr0MrWave:
**Issue:** Closes #257 · Refs #241

## What

- New `net.prompt_abandoned` audit event for a network prompt that was raised but never resolved.
- Prompt-sourced `net.decision` events now carry `waited_ms` — how long the dialog was open before it was answered.
- `omac diagnose` reports both, and no longer claims *"Nothing was blocked by the network policy"* for a run where either happened.

## Why

While a prompt is open the connection is held, and the requesting tool is subject to **its own** timeout. The opencode Skainet plugin aborts model discovery after 3s; answering a GUI dialog measurably takes longer (5s measured here, with the dialog already in focus). So the tool gives up, the fetch fails silently, and the allow you then grant applies to a request nobody is waiting for.

Both halves of that were invisible:

- If the run ends with the dialog still open, **no** `net.decision` is ever written. `omac diagnose` reported `0/1 connection(s) blocked · Nothing was blocked by the network policy` for a run that had silently failed.
- If the dialog *is* answered, just too late, a perfectly ordinary allow is recorded and nothing looks wrong at all.

This is not an edge case: the compiled-in default profile (`internal/sandboxprofile/resolve.go:23`) has no `allow_domain`, so on a fresh install every host prompts on first contact.

## How

- The dialog deliberately runs on a detached context (`internal/netprompt/prompt.go:312` builds its timeout from `context.Background()`), so an open prompt outlives the request that raised it and dies with the process. Detection therefore belongs at teardown: `Filter.DrainAbandonedPrompts` records whatever is still in flight when the proxy closes. It runs **last** in `Server.Close`, after the listener and connections are gone — draining earlier clears the coalescing map while requests can still arrive, which would raise a second dialog for a host already being asked about.
- Nothing in the proxy can observe the requester leaving (`clientSourceCtx` is also `context.Background()`-derived), so for the answered-too-late case the prompt's open duration is recorded on the decision instead, and diagnose flags anything ≥3s (`promptLatencyThresholdMS` — the shortest client budget observed in the wild).
- A prompt already recorded as abandoned does not also emit a decision: `promptWait` carries a `drained` flag so one request cannot appear as both abandoned and allowed.
- Abandoned entries are aggregated per `host:port` with a repeat count and capped at 8, mirroring `maxBlockedShown` — otherwise `--run all` emits one line per occurrence across the whole persistent log.

## Verification

`go build ./...`, `go vet ./internal/...` clean. `go test ./internal/netproxy/ -race` clean. `go test ./...` passes except `TestIntegrationWorktreeKnownLimitations` and `TestIntegrationWorkflowInterpretersRunnable`, which fail identically on clean `main` on this host (local toolchain paths absent from the default profile).

Drain path, end to end, **no human interaction** (the client times out on its own):

```
omac sandbox run --profile <prompt-enabled> --audit-log /tmp/a.jsonl \
  -- curl -s --max-time 1 https://example.com          # exit 28

diag:  net ABANDONED example.com:443 (prompt raised 967ms ago, no verdict —
       the requesting tool stopped waiting)
audit: net.prompt_abandoned example.com 443 waited_ms=967
omac diagnose:
       last run · filtered mode · 0/0 connection(s) blocked · 1 prompt(s) abandoned
       • 1 network prompt(s) … were raised but never answered in time
```

Late-answer path (the case the first commit missed — a `waited_ms=5200` prompt-sourced allow):

```
• 1 prompt(s) were answered after 3.0s or more — a short-timeout tool may have failed anyway
  chat.model.tngtech.com:443 was held for 5.2s before the prompt was answered.
  Fix: add these hosts to network.allow_domain so no prompt is raised.
```

…and no *"Nothing was blocked"* in either case. Runs with neither condition are unchanged (regression test asserts the note still fires).

The second commit is review fixes; the late-answer detection, the `Close` ordering, the double-report suppression and the detail cap each have a test.

## Follow-up

- **Correction to #257 as filed:** the issue implies a cancelled request context. The dialog is actually detached and dies with the process, so detection had to move to teardown plus a recorded wait. Worth a comment on the issue.
- Deliberately not included (#257 checklist item 4): making a bare `omac sandbox run` write the central audit trail. It only gets an auditor when the parent passes `--audit-log` (`internal/sandboxrun/run.go:144`), and wiring it would couple `sandboxrun` to launcher-config loading — worth its own change.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -170,7 +170,9 @@ omac cannot pass on your access token, so packages that require login still fail
 
 ## Audit trail
 
-omac logs every security-relevant action to an append-only file: process launches, network decisions, secret injections. The file is outside the sandbox so the agent cannot tamper with it.
+omac logs every security-relevant action to an append-only, structured (JSON Lines) file the agent cannot tamper with: the sandboxed inner command and each sidecar it spawns (`process.exec` / `process.exit`), outbound network allow/deny decisions and their source (`net.decision`), network prompts raised but never answered in time (`net.prompt_abandoned`), facade requests (`facade.request`), control-plane mutations (`control.mutation`), secret injection by name (`secret.inject`), non-ready routes (`route.state`), and session lifecycle (`session.start` / `session.stop`).
+
+A prompt-sourced `net.decision` also carries `waited_ms`, how long the dialog was open before it was answered. This distinguishes "allowed promptly" from "allowed after the requesting tool had already timed out" (see [Short-timeout clients](#short-timeout-clients)); `omac diagnose` flags the latter.
 
 | Platform | Default path |
 |---|---|
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -144,6 +144,12 @@ Java (Maven/Gradle) and Node/npm do not reliably route their package downloads t
 
 Node injection requires Node ≥ 22.21.0 (22.x line) or ≥ 24.5.0; on older versions it is skipped and downloads may still fail.
 
+### Short-timeout clients
+
+While a network prompt is open the connection is held, and the requesting tool is subject to *its own* timeout — some are only a few seconds (the opencode Skainet plugin aborts model discovery after 3s). Answering a dialog reliably takes longer than that, so the tool can give up before your click lands: the fetch fails silently, and the allow you then grant applies to a request nobody is waiting for any more. Put hosts that short-timeout tools depend on in `network.allow_domain` so no prompt is raised. Under `omac serve` or any non-TTY launch there is no dialog at all and `on_unavailable: deny` (the default) makes this deterministic.
+
+`omac diagnose` reports these as abandoned prompts — a run whose prompt was never answered in time no longer reads as "nothing was blocked".
+
 ### Private package registries
 
 If your company hosts its own npm packages, `~/.npmrc` says where to find them. A line like `@acme:registry=https://npm.acme.test` means "packages starting with `@acme/` come from that server".
--- a/internal/audit/audit_test.go
+++ b/internal/audit/audit_test.go
@@ -246,7 +246,7 @@ func TestFailOpenWarnsOnce(t *testing.T) {
 
 func TestValidJSONPerLine(t *testing.T) {
 	a, path := newTestAuditor(t, Config{})
-	a.Emit(NetDecision("example.com", 443, true, "host", "prompt", true))
+	a.Emit(NetDecision("example.com", 443, true, "host", "prompt", true, 0))
 	a.Emit(SecretInject("s", "tok", []string{"A", "B"}, []string{"C"}))
 	a.Emit(RouteStateEvent("s", "tok", "broken", "boom"))
 	_ = a.Close()
@@ -293,7 +293,7 @@ func TestInheritedRunIDPreservesCorrelation(t *testing.T) {
 	if err != nil {
 		t.Fatalf("child New: %v", err)
 	}
-	child.Emit(NetDecision("host.example", 443, true, "host", "prompt", false)) // seq 1
+	child.Emit(NetDecision("host.example", 443, true, "host", "prompt", false, 0)) // seq 1
 	_ = parent.Close()
 	_ = child.Close()
 
--- a/internal/audit/event.go
+++ b/internal/audit/event.go
@@ -34,15 +34,22 @@ const (
 
 // Event types. Dotted namespaces group related actions.
 const (
-	TypeSessionStart    = "session.start"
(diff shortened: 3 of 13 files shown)


mwtng:
I just rebased and reran the verification. I did not do a review, yet. 

But probably one should have a look at the changes to the documentation and if they are really necessary in such detail.

Issue #267, opened 2026-08-27
Support per-project sandbox grants


VictoriaRuckerbauer:
Context / Summary
A user wants different omac access grants per project (e.g. one project needs an extra open port, another needs a different filesystem grant). The launcher config already supports a project-local file (<project>/.opencode/oh-my-agentic-coder.yaml), soon documented under Configuration (update-readme branch) → Per-project 
configuration. But the sandbox grants file has no such per-project support.

Problem / What
LoadLauncher (internal/config/launcher.go:363) resolves the launcher config per workdir, but sandbox grants are resolved only from the user-global location: internal/sandboxprofile/resolve.go resolves a bare profile ref to ~/.config/omac/sandbox-profiles/<ref>.json, and the default launch always passes
--profile default (internal/config/launcher.go:212). A file at <project>/.opencode/sandbox-profiles/default.json is not auto-discovered, so per-project grants (paths, hosts, network.open_port) are currently impossible without an explicit --profile <path>.

Suggested fix / Ask
- [ ] Decide the desired precedence (workdir grants override global? merge or replace?)
- [ ] Add workdir-local discovery for the grants file, mirroring the launcher config's candidate order
- [ ] Surface it in omac doctor / provenance
- [ ] Document it and remove the "no per-project equivalent" caveat from docs/configuration.md

shleder:
Per-project sandbox policy is the shape [vetto](https://github.com/shleder/vetto) (my project) uses, as a data point for the design: every session takes a TOML policy (resolved from the project or CLI) defining read/write roots, deny paths, network mode (off / per-domain allowlist via a host-side DNS-validating broker), rlimits and a hard timeout — so "this project needs an extra port" is one line in the project-local policy, and the enforcement is kernel-level (Landlock + namespaces + seccomp on Linux) rather than config the agent can see or influence.

If omac wants OS-enforced grants behind the launcher config, wrapping the spawned agent in vetto would give the grants file a kernel backstop. Disclosure: I'm the author.

VictoriaRuckerbauer:
Idea: treat sandbox settings like Claude settings: if there is a local file in the project, the global file is overwritten.
However, the agent must not be able to access the sandbox settings (do not make it available within the sandbox)!

PR #275, opened 2026-09-01
Simplify sandbox config

### LINKED ISSUE #272: Remove outdated nono option
### Duplicate check

### Context / Summary

The nono sandbox support is a long deprecated feature and should be removed to simplify the application.

### Problem / What

Nono sandbox is outdated.

### Suggested fix / Ask

Remove nono support.

### Non-goals

_No response_

### Evidence / Environment

_No response_

VictoriaRuckerbauer:
**Issue:** Closes #272

## What
Remove options for using any but the built-in sandbox (which is the heart of omac and, therefore, should be used anyway).

## Why
Nono-support cluttered the code and the config. This simplifies both.

## Follow-up
- After some time, delete the detection-only fields (SandboxConfig.DefaultProfile, .Profiles) and validateSandbox itself. They exist only to catch and reject old configs; once users have migrated, the whole migration layer can be removed.


VictoriaRuckerbauer:
Current status:
- The launcher config can be specified on a per-project basis.
- In the launcher config, you can define a profile for sandbox grants (relative or absolute path).
- Sandbox grants are not git-ignored by default; network settings are (they likely change too frequently to track in VCS).
- When a profile is set, network settings are placed as siblings of the profile path. Feel free to discuss whether this is the right approach.

With this, the PR should now also satisfy #267. However, I did not want to split them in order to have only one breaking change for user-facing settings.

I have not yet had time to do some tests of this in daily usage.

mwtng:
**Status update:**

**Layered profile system:** 
Launcher config's `sandbox.profile_name` (renamed from `profile_path`) selects by file name inside exactly one layer: 
- `~/.config/omac/sandbox-profiles/` (user-global) 
- `<workdir>/.omac/` (project) 

Layers never mix, either full global or full local. The old `sandbox.default_profile`/`profiles` keys are rejected with a migration hint.

**Protected project layer:** 
- `omac start`/`serve` create `<workdir>/.omac/` (also under `--no-sandbox`), to prevent agent-planted configs for next session
- the directory is masked as read-only inside the session with an explanatory `.omac-denied` marker
- the `<profile>.json`, its `<profile>.pages.json` sibling, and the launcher `config.yaml` in `.omac/` are not accessible to the agent
- symlinked paths are refused

**New CLI argument:**
`--profile-path` allows specifying a profile to be selected (bypasses the launcher `config.yaml`). The file must lie in either global or local layer (to be protected from agent-editing).

**Runtime integrity pin:**
- approved project-local content is hashed into `~/.config/omac/project-sandbox.json` (host-only, invisible to the session)
- if the `.omac` content changed since approval, the next launch **refuses to start** until it is re-approved with `--accept-project-config`. 
- the gate covers an explicit `--profile-path` into the project layer too
- a global-layer path needs no approval (`~/.config/omac` is user-maintained and never forwarded into the sandbox)

MacOS note: Seatbelt cannot prevent directory replacement inside the workdir. This is exactly why the anchor exists and is host-side.

**Validation suggestions** (build first: `go build -o ./omac ./cmd/omac`):

1. `./omac doctor` / `./omac start <harness>` with a config from the latest release → legacy keys are rejected
   with migration hints; fix them (after a backup) until the local `omac` runs.
2. Fresh repo → `.omac/` is created; the agent sees the directory but cannot read or write it.
3. Add `.omac/default.json` (e.g. with content `{}`), start 
  → next start aborts with hint 
  → next start with `--accept-project-config` pins it (and shows a warning if the content was empty)
  → next start (without `--accept-project-config`) starts immediately
4. `--profile-path` into either layer works (first use automatically pins unpinned file)
    - outside → rejected
    - symlink → refused
    - a project-layer path behaves like the config layer on tampering (abort + re-approve)
    - a global-layer path is always admitted
5. Same checks with `./omac serve`.

Issue #277, opened 2026-09-01
Publish docs/ as a documentation site on GitHub Pages


nhuelstng:
## Context / Summary

The docs under `docs/` are already written for a static site generator: every file carries `title`/`description` frontmatter, `docs/README.md` reads as a landing page, and three files use Docusaurus admonitions (`docs/advanced/nono.md:6`, `docs/advanced/serve-mode.md:6`, `docs/contributing/serve-spec.md:6`). Nothing renders them today, so readers get raw markdown on GitHub.

## Problem / What

- `:::warning[Deprecated]` and friends show up as literal text on github.com.
- The mermaid architecture diagram in `docs/README.md:20` is not rendered by GitHub's markdown in all views and has no styling.
- There is no navigation, no search, and no single URL to point users at. `README.md:41` links into the docs with hard-coded `github.com/.../tree/main/...` URLs.

## Suggested fix / Ask

Add a Docusaurus site that consumes `docs/` in place, plus a GitHub Pages deployment workflow.

- [x] `website/` Docusaurus project reading `../docs` in docs-only mode (`routeBasePath: '/'`), so no doc file moves
- [x] Mermaid via `@docusaurus/theme-mermaid`
- [x] Manual sidebar in `website/sidebars.js`
- [x] `.github/workflows/docs.yml`: build on PRs, deploy on `main`, `workflow_dispatch` for branch previews
- [x] Fix the one MDX-incompatible construct: bare autolink `docs/troubleshooting.md:140`
- [x] Point `README.md` at the published site instead of `tree/main` URLs

## Non-goals

- Moving the root `README.md` into the site. It stays repo-facing (badges, contributing, license); `docs/README.md` is the site home.
- Versioned docs, i18n, blog, custom branding.
- Removing markdown readability on GitHub. `docs/` must keep working as plain files in the repo.

## Evidence / Environment

Docusaurus flavour already in the docs:

```
$ grep -rc '^:::' docs --include='*.md' | grep -v ':0'
docs/advanced/nono.md:2
docs/contributing/serve-spec.md:2
docs/advanced/serve-mode.md:2
```

🤖 Filed with Claude Code

mwtng:
Closed by #278 
Documentation published here: https://tng.github.io/oh-my-agentic-coder/

PR #278, opened 2026-09-01
docs: publish docs/ as a Docusaurus site on GitHub Pages

nhuelstng:
**Issue:** Refs #277

## What

- `website/` — Docusaurus project that reads `../docs` in place (docs-only mode, `routeBasePath: '/'`). No doc file is moved or copied, so `docs/` stays readable as plain markdown on GitHub.
- `.github/workflows/docs.yml` — build on PRs (never deploys), build + deploy on `main`, and `workflow_dispatch` for previewing a branch at the real URL before merge.
- Three link fixes in `docs/` that MDX cannot compile, and a "Documentation site" section in `docs/contributing/development.md`.

## Why

The docs were already written for Docusaurus — `title`/`description` frontmatter on all 16 files, `:::warning[Deprecated]` admonitions in three of them, a mermaid diagram in `docs/README.md`. On github.com the admonitions show up as literal `:::` text and there is no navigation, search, or single URL to point users at. This renders what is already there rather than rewriting it.

## How

- `docs/README.md` becomes the site home: Docusaurus treats `README.md` as its folder's index, and `routeBasePath: '/'` puts it at the root. The root `README.md` stays repo-facing and now links to the published site.
- Sidebar is hand-ordered in `website/sidebars.js` (overview → install → daily use → reference). Autogenerating it would need a `_category_.json` in every subfolder and would sort alphabetically.
- `onBrokenLinks: 'throw'` plus `markdown.hooks.onBrokenMarkdownLinks: 'throw'`, so a dead internal link fails the build. That is what caught the three link fixes:
  - `docs/troubleshooting.md` — bare autolink `<https://…>` is a JSX parse error under MDX 3.
  - `docs/troubleshooting.md` → `../COLLABORATION.md`, `docs/contributing/development.md` → `../../.goreleaser.yaml` and `../../.github/workflows/release.yml`: files outside `docs/` have no page to resolve to, so they are now full GitHub URLs.
- Mermaid via `@docusaurus/theme-mermaid` (+ `@mermaid-js/layout-elk`, which it requires at build time).
- `baseUrl: '/oh-my-agentic-coder/'` works unchanged on a fork's Pages site, since the URL path is the repo name.

## Verification

```
$ cd website && npm ci && npm run build
[SUCCESS] Generated static files in "build".

$ find build -name index.html | wc -l
16                      # every docs/*.md has a page; no broken links (build throws on them)

$ grep -o 'theme-admonition-warning' build/advanced/nono/index.html
theme-admonition-warning                 # ::: blocks render as admonitions

$ grep -rl ':::warning\|:::caution\|:::note' build --include='*.html'
                        # no literal ::: left in the output

$ grep -o '<title[^>]*>[^<]*</title>' build/security/index.html
<title data-rh="true">Security model | oh-my-agentic-coder</title>   # frontmatter used
```

Mermaid renders client-side, so the diagram is not in the SSR'd HTML (its source is in `build/assets/js/`). It still needs one visual check in the preview deployment.

## Follow-up

Two things a repo admin has to do before this URL is live, neither of them in the diff:

1. Settings → Pages → **Source: GitHub Actions**.
2. To preview from this branch before merging: Settings → Environments → `github-pages` → allow `docs/github-pages` under deployment branches, then run the **Docs** workflow via `workflow_dispatch` on this branch. Nothing is published at that URL yet, so a branch deploy overwrites nothing.


nhuelstng:
You can a have look how this page will look like on the github pages of my fork for that purpose:
https://nhuelstng.github.io/oh-my-agentic-coder/

If we want to deploy from another branch than main, @thomas-endres-tng needs to add that setting

VictoriaRuckerbauer:
I think there is some formatting issue for troubleshooting.md (probably the first line being empty).

website/sidebars.js
```diff
@@ -0,0 +1,46 @@
...
+      collapsed: false,
+      items: ['getting-started/quick-start'],
+    },
+    {
+      type: 'category',
+      label: 'Usage',
+      collapsed: false,
+      items: ['usage/cli', 'usage/skills', 'configuration'],
+    },
+    'security',
```
VictoriaRuckerbauer: Personal opinion: I think that these should be the last two pages.

website/sidebars.js
```diff
@@ -0,0 +1,46 @@
...
+    'troubleshooting',
+    {
+      type: 'category',
+      label: 'Skills',
+      items: ['skills/authoring'],
+    },
+    {
+      type: 'category',
+      label: 'Advanced',
+      items: ['advanced/cache', 'advanced/serve-mode', 'advanced/nono'],
```
VictoriaRuckerbauer, line 28: I'm currently working on a PR to remove nono, which will also remove this file. If https://github.com/TNG/oh-my-agentic-coder/pull/275 is merged before this PR, the line needs to be adjusted.

website/docusaurus.config.js
```diff
@@ -0,0 +1,89 @@
...
+        theme: {
+          customCss: './src/css/custom.css',
+        },
+      },
+    ],
+  ],
+
+  themeConfig: {
+    navbar: {
+      title: 'omac',
```
VictoriaRuckerbauer, line 55: Not sure whether it belongs here:
The UI looks like there should be two buttons in the header (omac and docs, both converting the mouse into click-visualization) which are not clickable. Imo, the header should go completely.

docs/troubleshooting.md
```diff
@@ -150,5 +150,5 @@ Before opening one:
    version, and the harness you were running (for example `claude-code` on
    Ubuntu 24.04).
 
-Keep it concise. The [collaboration guide](../COLLABORATION.md#issue-rules) has
+Keep it concise. The [collaboration guide](https://github.com/TNG/oh-my-agentic-coder/blob/main/COLLABORATION.md#issue-rules) has
```
VictoriaRuckerbauer, line 152: Should we also link docusaurus here instead?

docs/troubleshooting.md
```diff
@@ -137,7 +137,7 @@ When you answer the network prompt, "session" decisions are held in memory only,
 
 If `omac doctor` and the fixes above don't resolve your problem, please report it
 as a GitHub issue at
-<https://github.com/TNG/oh-my-agentic-coder/issues>.
+[github.com/TNG/oh-my-agentic-coder/issues](https://github.com/TNG/oh-my-agentic-coder/issues).
```
VictoriaRuckerbauer, line 139: Should we also link docusaurus here instead?

docs/contributing/development.md
```diff
@@ -33,12 +33,33 @@ go build -trimpath -ldflags "-s -w -X main.Version=0.1.0-local" -o omac ./cmd/om
 ./omac version   # -> omac 0.1.0-local
 ```
 
-For full multi-platform artifacts (`.deb`, `.pkg.tar.zst`, checksums), use GoReleaser (`goreleaser release --clean --snapshot --skip=publish`). See [`.goreleaser.yaml`](../../.goreleaser.yaml).
+For full multi-platform artifacts (`.deb`, `.pkg.tar.zst`, checksums), use GoReleaser (`goreleaser release --clean --snapshot --skip=publish`). See [`.goreleaser.yaml`](https://github.com/TNG/oh-my-agentic-coder/blob/main/.goreleaser.yaml).
+
+## Documentation site
```
VictoriaRuckerbauer, line 38: I'd argue that this bloats the documentation.
Imo, we should only keep the following information:
- we are using docusaurus
- how to deploy it

.github/workflows/docs-pages.yml
```diff
@@ -0,0 +1,80 @@
+# Documentation site (Docusaurus over the repo's docs/ directory) → GitHub Pages.
+#
+# Three triggers, on purpose:
+#   pull_request      — build only. A dead link or an MDX parse error fails the
+#                       PR instead of reaching the published site.
+#   push to main      — build and deploy. This is the live site.
```
VictoriaRuckerbauer: I don't think that we allow pushes to main?
Does this even make sense?

.github/workflows/docs-pages.yml
```diff
@@ -0,0 +1,80 @@
...
+#   workflow_dispatch — build and deploy from any branch, so a docs change can be
```
VictoriaRuckerbauer: I don't think that this clearly tells the reader when the documentation is updated.
What exactly do I need to do?
Also, this information should probably be in `development.md` (additionally).

.github/workflows/docs-pages.yml
```diff
@@ -0,0 +1,80 @@
...
+    # Never on pull_request: a PR from a fork must not be able to publish.
+    if: github.event_name != 'pull_request'
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
VictoriaRuckerbauer, line 85: I have probably missed something, but where is this defined?

VictoriaRuckerbauer (changes requested):

.github/workflows/docs-pages.yml
nhuelstng: as I know github actions "push to main" does not necessarily mean a direct push but also a merge; probably we only want to keep this as the only trigger (

.github/workflows/docs-pages.yml
nhuelstng: workflow dispatch should only be a fallback for quick manual fixes; I would prefer that everything that lands on main is then reflected there

.github/workflows/docs-pages.yml
nhuelstng: directly below this comment
some things like url are set by the deploy pages flow

docs/troubleshooting.md
nhuelstng: yes, that makes sense

docs/contributing/development.md
nhuelstng: "how to deploy it" -> only the actions flow (no manual?)

website/sidebars.js
nhuelstng: I would say still before contributing, but open to discuss that

docs/troubleshooting.md
mwtng: For that, we probably would have to move that file into the `docs/` directory as well, right? 

Probably rather something for a separate Issue/PR. Maybe a symbolic link from docs to the root dir could be helpful for people to quickly find it (and have a single source of truth). But there should probably happen some merging then from the current docs content and this file.

mwtng (approved):

PR #255, opened 2026-08-22
fix: correct module path to github.com/TNG/oh-my-agentic-coder

CodingTil:
**Issue:** Closes #98 Finding 1
#99 does not truly solve this finding, it only mitigates it.

## What
* Corrects the go module path to the actual git remote
* This enables installation via `go install github.com/TNG/oh-my-agentic-coder/cmd/omac@latest`

## Why
* Simple installation option
* Good practice to point the go module path to the actual git remote

## How
* Renamed all findings of `github.com/tngtech/oh-my-agentic-coder` to `github.com/TNG/oh-my-agentic-coder`
* Adjusted documentation in `docs/INSTALLATION.md`

## Verification
* `go test ./...`
* `go vet ./...`
* `gofmt -l .`
* `go build cmd/omac/main.go`

## Follow-up
-/-

PR #280, opened 2026-09-02
Fix/module path tng

VictoriaRuckerbauer:
**Issue:** replaces #255

## What
Correct URL references in repo to make `go install` work.

## Verify
`GOPROXY=direct GOSUMDB=off GOBIN=/tmp/omac-test \
 go install github.com/TNG/oh-my-agentic-coder/cmd/omac@fix/module-path-tng`


mwtng (approved):
Checked after rebase on main that still all imports are adjusted

Issue #281, opened 2026-09-08
Security Scan workflow has failed every week since it was added: archive push refused, and the scan itself is a silent no-op


nhuelstng:
## Context / Summary

Every scheduled run of `.github/workflows/security-scan.yml` has failed. The four
runs in the retained history all fail the same way, in the same step:

| Run | Date | Result |
|---|---|---|
| [31873510508](https://github.com/TNG/oh-my-agentic-coder/actions/runs/31873510508) | 2026-08-15 | failure |
| [32561281623](https://github.com/TNG/oh-my-agentic-coder/actions/runs/32561281623) | 2026-08-22 | failure |
| [33242172274](https://github.com/TNG/oh-my-agentic-coder/actions/runs/33242172274) | 2026-08-29 | failure |
| [33954142544](https://github.com/TNG/oh-my-agentic-coder/actions/runs/33954142544) | 2026-09-05 | failure |

There are two independent problems, and the first one hides the second.

### 1. The archive push is refused (this is the red X)

`.github/workflows/security-scan.yml:246` clones and pushes the private archive
repo with `SECURITY_SCAN_PAT`. Every run ends:

```
remote: Write access to repository not granted.
fatal: unable to access 'https://github.com/***.git/': The requested URL returned error: 403
##[error]Process completed with exit code 128.
```

The token authenticates (no 401) but is not allowed to write the archive repo.
This is a secret/permission problem, not a code one — but the workflow reports it
as a bare `exit code 128` against a redacted URL, which reads like "the security
scan failed" rather than "fix the PAT". The same secret is used by
`.github/workflows/e2e-smoke.yml:1166`, whose `Record & report` job also fails.

### 2. The scan itself never actually runs, and reports success anyway

In all four runs the `deep` scan finishes in ~105 seconds and writes no
`vulnerabilities.json`:

```
strix_runs/target_36ca//vulnerabilities.json not written -- treating as zero findings
```

A 105-second `--scan-mode deep` pentest is not a findings-free scan, it is a scan
that died on startup. It is reported as a clean pass because:

- `.github/workflows/security-scan.yml:171` discards strix's exit code with
  `|| true`, so a crash is invisible;
- the only completeness guards are "a run directory exists"
  (`security-scan.yml:186`) and "the log has no `context_length_exceeded`"
  (`security-scan.yml:203`), both of which a crashed run satisfies;
- `.github/scripts/security_scan_triage.py:74` then treats the missing findings
  file as zero findings.

[...]

PR #282, opened 2026-09-08
fix(security-scan): preflight the archive credential and stop hiding a no-op scan

nhuelstng:
**Issue:** Refs #281

## What

- Probe the archive credential in the first ten seconds of the job instead of discovering it is read-only three hours later, and name the permission to grant.
- Keep strix's exit code instead of discarding it with `|| true`, so a scan that dies on startup can no longer pass as a clean one.
- Put the scan's exit code, duration and log size on the public step summary.
- Pin the scan's **own** model, at `zai-org/GLM-5.3` — it had none, and used a different one on a scheduled run than on a manual one.
- Stop the job timeout from **destroying** a scan's results: strix gets its own budget, and the archive rediscovers the run directory instead of trusting a step output the failure path never writes.
- Scan a symlink-free copy of the checkout — `strix --target` refuses any tree containing a symlink, which is why the scan has never actually run. See the comment thread for the root cause.

Does **not** fix the token itself — that is a secret change outside this repo. See Follow-up.

## Why

Every scheduled run since the workflow was added has failed, for two reasons the job reported as one opaque exit code:

| Run | Date | Scan duration | Result |
|---|---|---|---|
| [31873510508](https://github.com/TNG/oh-my-agentic-coder/actions/runs/31873510508) | 2026-08-15 | 105s | failure |
| [32561281623](https://github.com/TNG/oh-my-agentic-coder/actions/runs/32561281623) | 2026-08-22 | 109s | failure |
| [33242172274](https://github.com/TNG/oh-my-agentic-coder/actions/runs/33242172274) | 2026-08-29 | 103s | failure |
| [33954142544](https://github.com/TNG/oh-my-agentic-coder/actions/runs/33954142544) | 2026-09-05 | 111s | failure |

**The red X** is the archive push. `SECURITY_SCAN_PAT` can read the private archive repo but not write it:

```
remote: Write access to repository not granted.
fatal: unable to access 'https://github.com/***.git/': The requested URL returned error: 403
##[error]Process completed with exit code 128.
```

**Underneath it**, the scan never ran. A `--scan-mode deep` pentest finishing in ~105 seconds with no `vulnerabilities.json` is not a findings-free scan. It passed anyway because `|| true` threw away the exit code, and the two completeness guards — a run directory exists, the log has no `context_length_exceeded` — are both satisfied by a process that died on startup.

So fixing the token on its own would have turned the job **green while it scanned nothing**, which is worse than the red X. Both had to move together.

The log that would explain the short scan is deliberately never printed publicly, and its only other destination is the archive repo the push fails on — so it has never been written anywhere. Breaking that circularity is the point of the preflight.

## How

- **`Preflight archive credential`** (new step, after checkout): asks for the `git-receive-pack` ref advertisement. GitHub authorizes that advertisement, so it is a read-only probe of *write* access — 200 writable, 403 read-only, 401 expired, 404 invisible. `git ls-remote` cannot tell those apart; it only ever exercises upload-pack. The token goes in `curl -u`, not the URL, so it stays out of anything curl prints.
- **`Run strix scan`**: `|| true` → `|| rc=$?`, plus `$SECONDS` around the call. Fails when strix exited non-zero **and** wrote no findings file — no evidence of a result at all. A findings-free scan that exits 0 still passes, preserving #194.
- **`Push full report`**: gated on the preflight having passed, so a bad token yields one actionable error rather than two, and `git commit`'s nothing-to-commit is no longer reported as an archiving failure. An `ERR` trap explains a push that fails *after* the preflight said it would not (protected branch, mid-run revocation).
- Disclosure model unchanged: exit code, duration and log byte count are numbers about the run, not about what it found. The header comment says so explicitly.

## Added: the scan now pins its own model, at GLM 5.3

Folded in here rather than onto a branch of its own: it is a second concern, but it edits the same file and a parallel branch would only conflict.

The scan had **no model of its own**, and ran a different one depending on how it started:

| Trigger | `E2E_MODEL` | Model actually used |
|---|---|---|
| schedule (weekly) | empty | `zai-org/GLM-5.2` — opencode's *harness* pin in `internal/e2e/versions.go` |
| workflow_dispatch | input default | `deepseek-ai/DeepSeek-V4-Flash` — the head of the **fallback** chain |

So the weekly scan silently tracked whatever the harness matrix happened to be tested against, and a manual scan was not comparable to it. Both now resolve to `zai-org/GLM-5.3`, which the gateway advertises.

`internal/e2e/versions.go` is deliberately **not** touched, so no e2e leg changes model. Resolution is otherwise unchanged: `E2E_MODEL_OPENCODE` still wins, the `-TEE` variant flip and the fallback chain still apply, and a value typed into the dispatch form still overrides the pin for that run. The triage step follows the same model via `STRIX_LLM`, so it moves too.

```
$ E2E_MODEL='zai-org/GLM-5.3' scripts/resolve-model.sh opencode
zai-org/GLM-5.3
$ E2E_MODEL='zai-org/GLM-5.3' scripts/probe-model.sh opencode      # no gateway creds -> passthrough
probe-model: no gateway base URL / token on the env - using 'zai-org/GLM-5.3' unprobed
zai-org/GLM-5.3
$ env -u E2E_MODEL scripts/resolve-model.sh opencode               # harness pin untouched
zai-org/GLM-5.2
$ E2E_MODEL='tngtech/DeepSeek-TNG-R1T2-Chimera' scripts/resolve-model.sh opencode
tngtech/DeepSeek-TNG-R1T2-Chimera
$ go test -tags e2e_fast ./internal/e2e/ -run 'Model|Version|Candidate|Fallback' -count=1
ok      github.com/TNG/oh-my-agentic-coder/internal/e2e 0.007s
```

## Added: a scan that runs out of time now keeps what it found

The previous commit got strix past startup, and the first real `deep` run
([34258819205](https://github.com/TNG/oh-my-agentic-coder/actions/runs/34258819205))
answered the open question in Follow-up below: **the 180-minute budget does not
hold.** It also exposed something worse than the overrun.

That run reached **22 findings** and archived **none** of them.

| | |
|---|---|
| Job | `17:43:38Z → 20:44:39Z`, exactly `timeout-minutes: 180`, conclusion `cancelled` |
| Step 8 `Run strix scan` | `cancelled` mid-scan |
| Steps 9–10 triage/summary | `skipped` |
| Step 12 `Push full report` | `success` — and pushed `strix_scan.log` + `proxy.log` **only** |

Two independent causes:

**1. The job's own ceiling ended the scan.** `timeout-minutes: 180` is enforced
by GitHub cancelling the job, which is the worst possible way for a long scan to
end: the step never finishes, every later step is skipped or reads outputs it
never wrote, and the runner is torn down with the run directory still on it.

strix now runs under `timeout -s INT -k 120s "${budget}m"`, sized by a new
`scan_budget_minutes` input (default 300), with `timeout-minutes` raised to 350
— just under the 360-minute hosted-runner ceiling — so the job's ceiling is no
longer the thing that ends a scan. An overrun becomes an ordinary exit code
*inside* the step, and the guards, triage and archive below all still run.
`-s INT` rather than the default `TERM` because SIGINT is the signal an
interactive strix already handles, so it gets a chance to flush; `-k` gives it
two minutes before SIGKILL.

**2. The archive could only find the run directory through a step output.**
`run_dir` is written to `$GITHUB_OUTPUT` *after* strix returns, so every way the
scan step can end early leaves it empty — and `Push full report` then copied
nothing while reporting success. The directory was on disk the whole time. It
now rediscovers `strix_runs/*/` itself and warns that it had to. **That one
change alone would have saved the lost run.**

### Partial is reported as partial, not as failure

A three-hour scan cut off with 22 findings was filed as `-FAILED` next to runs
that died in 105 seconds. Now:

- a third archive status, `PARTIAL`, distinct from `ok` and `FAILED`;
- triage gated on a findings file *existing* rather than on the scan step
  succeeding, so a partial run still gets a triage report;
- the **finding count** parsed off the log's own `╭─ VULN-NNNN` banners. strix
  writes `vulnerabilities.json` only at the end, but the log carries every
  finding it reported in full — that count is the one thing that separates
  "PARTIAL, 22 findings to read" from "achieved nothing". It is a number about
  the run, not a finding, so it is safe on the public summary; the header
  comment says so.

A budget stop that found *nothing* still fails, exactly as before — that is
indistinguishable from a scan that never got going, which is the regression the
first commit exists to prevent.

### Verification

`run:` blocks all parse (`bash -n`, `scripts/check-workflow-shell.py`: *8
workflow file(s): all run blocks are valid bash*). The scan guards and the
archive step were extracted and driven against stubs under
`bash --noprofile --norc -e -o pipefail` — GitHub's own shell flags, which the
earlier round of stub testing did not use:

```
case=overrun budget=2s   rc=0  partial=true   exit=124  vulns=3  have=false
       ::notice title=Partial scan kept
       ::warning title=Scan stopped at its budget
case=quiet   budget=2s   rc=1  partial=true   exit=124  vulns=0  have=false
       ::error::strix exited 124 ... and wrote no vulnerabilities.json
case=clean   budget=5s   rc=0  partial=false  exit=0    vulns=1  have=true
case=crash   budget=5s   rc=1  partial=false  exit=3    vulns=0  have=false
case=clean   budget=5x   rc=1  ::error::scan_budget_minutes must be a whole number
```

Archive step, with a stub `git`/`mktemp` and a run directory on disk:

```
outcome=cancelled  run_dir output=''   -> ::warning title=Recovered the run directory
                                          20260908-174410-manual-FAILED/
                                          {run.json,vulnerabilities.json,strix_scan.log,proxy.log}
outcome=success    partial=false       -> 20260909-120000-manual-ok
outcome=success    partial=true        -> 20260909-130000-manual-PARTIAL
outcome=cancelled  run_label=''        -> falls back to a timestamp
```

The first line is the 2026-09-08 shape: same inputs, results kept instead of
discarded.

## Verification

`strix` and the archive repo cannot be reached from a dev machine, so each `run:` block was extracted from the YAML and driven against stubs.

**YAML and shell parse** — every `run:` block, GHA expressions stubbed:

```
$ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/security-scan.yml'))"   # ok
$ bash -n <each run: block>
OK Preflight archive credential / Install strix / Start context-compaction proxy
OK Resolve and verify model / Run strix scan / Dedupe against open issues
OK Public summary / Stop context-compaction proxy / Push full report
```

**Preflight**, against a stub `curl` returning each status:

```
[200] rc=0 :: archive repo is writable
[403] rc=1 :: ::error title=Archive repo is read-only::...Grant the token Contents: read and write...
[401] rc=1 :: ::error title=Archive credential rejected::...expired or malformed...
[404] rc=1 :: ::error title=Archive repo not reachable::...
[000] rc=1 :: ::error title=Archive preflight inconclusive::...
[unset] rc=0 :: ::warning::SECURITY_ARCHIVE_REPO or SECURITY_SCAN_PAT secret missing
```

**Scan guards**, against a stub `strix` (this is the regression that mattered — case 1 is today's green no-op, case 2 is what it should have been):

```
1. exit 0, run dir, no findings   -> rc=0  PASS   (findings-free run, #194)
2. exit 1, run dir, no findings   -> rc=1  FAIL   ::error::strix exited 1 after 0s and wrote no vulnerabilities.json
3. exit 1, run dir, WITH findings -> rc=0  PASS   (findings not discarded over a non-zero exit)
4. exit 0, no run dir             -> rc=1  FAIL   ::error::strix produced no run directory
5. context_length_exceeded in log -> rc=1  FAIL   ::error::strix hit context_length_exceeded
```

Cases 4 and 5 are the pre-existing guards, confirmed still firing.

**Push step**, against a local bare repo standing in for the archive:

```
a. happy path            -> rc=0, archived: scans/20260908-120000-scheduled-ok/{agent.log,strix_scan.log,triage-report.md}
b. push refused (chmod)  -> rc=1, ::error title=Report archiving failed::The scan itself is not what failed here...
c. nothing staged        -> rc=0, ::warning::nothing to archive for this run
```

Also confirmed that bash's `ERR` trap keeps `set -e`'s exemption for the `[ -f x ] && cp` idiom the step already used, so the trap does not misfire on an absent `triage-report.md` or `proxy.log`.

## Follow-up

- `SECURITY_SCAN_PAT` **has been rotated and works** — the preflight passed and `strix_scan.log` reached the archive repo, which is how the root cause below was found.
- `.github/workflows/e2e-smoke.yml:1166` uses the same secret; its `Record & report` job fails the same way ([33954821483](https://github.com/TNG/oh-my-agentic-coder/actions/runs/33954821483)) and is fixed by the same token change.
- ~~Why strix dies after ~105 seconds is still unknown~~ **Answered and fixed in this PR** — `strix --target` aborts on the repo's `CLAUDE.md -> AGENTS.md` symlink, added two days after this workflow, so the scan has never once run. Root cause and verification in the comment thread.
- ~~Still unknown: whether the 180-minute budget and the compaction proxy hold up under a full scan~~ **The budget did not** — run [34258819205](https://github.com/TNG/oh-my-agentic-coder/actions/runs/34258819205) was cancelled at exactly 3h with 22 findings reported; addressed above. The compaction proxy held: no `context_length_exceeded` in that run's log.
- Still unknown: whether a scan reaches the end inside the new 300-minute budget, and what a *complete* run reports. The 22 findings from the cut-off run have been verified separately (19 stand, 2 rejected as by-design, 1 disputed) in the private archive repo.
- #281 could not be self-assigned — the sidecar exposes no assignee route.

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -7,11 +7,13 @@
 # redirected straight to a file and never printed to the console, because
 # this is a public repo and both the job log and (if used) upload-artifact
 # would be publicly downloadable. Full reports are pushed to a private
-# companion repo instead, and only aggregate severity/dedupe counts reach
-# the public step summary. A human reviews the private repo and decides
-# what -- if anything -- becomes a public GitHub issue. The private artifact
-# repo is named by the SECURITY_ARCHIVE_REPO secret (kept out of public
-# source); see it for the artifact history.
+# companion repo instead, and only aggregate severity/dedupe counts -- plus
+# the scan's exit code, duration and log size, which are numbers about the
+# run rather than anything it found -- reach the public step summary. A human
+# reviews the private repo and decides what -- if anything -- becomes a public
+# GitHub issue. The private artifact repo is named by the
+# SECURITY_ARCHIVE_REPO secret (kept out of public source); see it for the
+# artifact history.
 name: Security Scan
 
 on:
@@ -60,6 +62,56 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: target
+
+      - name: Preflight archive credential
+        id: archive
+        # The whole point of this job is the report it pushes to the private
+        # archive repo, so a credential that cannot write there makes the scan
+        # below wasted compute -- and until 2026-09 it failed exactly that way
+        # every week, three hours in, with nothing but "exit code 128" against a
+        # redacted URL to explain it. Check first, and say what to fix.
+        #
+        # GitHub authorizes the ref advertisement, so asking for the
+        # git-receive-pack advertisement is a read-only probe of *write* access:
+        # 200 means the push will be allowed, 403 means the token is valid but
+        # read-only on that repo, 401 means it is expired or malformed, and 404
+        # means it cannot see the repo at all. `git ls-remote` cannot tell these
+        # apart -- it only ever exercises upload-pack.
+        env:
+          SECURITY_SCAN_PAT: ${{ secrets.SECURITY_SCAN_PAT }}
+          ARCHIVE_REPO: ${{ secrets.SECURITY_ARCHIVE_REPO }}
+        run: |
+          set -euo pipefail
+          if [ -z "${ARCHIVE_REPO:-}" ] || [ -z "${SECURITY_SCAN_PAT:-}" ]; then
+            echo "::warning::SECURITY_ARCHIVE_REPO or SECURITY_SCAN_PAT secret missing — the scan will run but its report cannot be archived"
+            exit 0
+          fi
+          # -u keeps the token out of the URL, so it stays out of any error curl
+          # prints. Only the status code is captured.
+          code=$(curl -sS -o /dev/null -w '%{http_code}' \
+            -u "x-access-token:${SECURITY_SCAN_PAT}" \
+            "https://github.com/${ARCHIVE_REPO}.git/info/refs?service=git-receive-pack" || echo 000)
+          case "$code" in
+            200)
+              echo "archive repo is writable"
+              ;;
+            403)
+              echo "::error title=Archive repo is read-only::SECURITY_SCAN_PAT can read the private archive repo but not write it, so this scan's report has nowhere to go. Grant the token Contents: read and write on the archive repo (a fine-grained PAT also has to list that repo explicitly), then re-run. Same secret, same symptom in e2e-smoke.yml's Record & report job."
+              exit 1
+              ;;
+            401)
+              echo "::error title=Archive credential rejected::SECURITY_SCAN_PAT was refused by GitHub (401) -- it is expired or malformed. Rotate it and re-run."
+              exit 1
+              ;;
+            404)
+              echo "::error title=Archive repo not reachable::SECURITY_SCAN_PAT cannot see the repo named by SECURITY_ARCHIVE_REPO (404). Either the repo name is wrong or the token has no access to it at all."
+              exit 1
+              ;;
+            *)
+              echo "::error title=Archive preflight inconclusive::Probing the archive repo returned HTTP ${code}. Not starting a three-hour scan whose report may be undeliverable."
+              exit 1
+              ;;
+          esac
 
       - name: Set up Python
         uses: actions/setup-python@v5
@@ -166,9 +218,21 @@ jobs:
           # Redirected to a file, NEVER printed to this (public) job log --
           # strix's console output includes full vulnerability detail and
           # working PoCs once findings surface.
+          #
+          # The exit code is recorded rather than discarded. It used to be
+          # thrown away with `|| true`, which meant a strix that died on
+          # startup was indistinguishable from one that ran for hours and found
+          # nothing -- every scheduled run from 2026-08-15 to 2026-09-05
+          # "succeeded" after ~105 seconds without writing a findings file. It
+          # is not asserted on here on its own, because a non-zero exit from an
+          # otherwise complete scan should not discard the findings; the check
+          # below combines it with whether a result was actually produced.
+          started=$SECONDS
+          rc=0
           strix -n --target "$PWD/target" \
             --scan-mode "${{ inputs.scan_mode || 'deep' }}" \
-            > strix_scan.log 2>&1 || true
+            > strix_scan.log 2>&1 || rc=$?
+          elapsed=$(( SECONDS - started ))
 
           # strix_runs/ lands relative to strix's OWN invocation cwd (this
           # job's default workspace root), as a sibling of target/ -- NOT
@@ -201,6 +265,28 @@ jobs:
             exit 1
           fi
 
+          # How long the scan ran and how much it wrote. Both are counts, not
+          # content, so they are safe on this public summary -- and they are the
+          # cheapest way to spot a scan that did not really happen: a `deep`
+          # scan of this repo takes hours and writes megabytes, so two minutes
+          # and a few kilobytes means something went wrong even when every
+          # guard above is satisfied. Published here because the log that would
+          # explain it only ever reaches the private archive repo.
+          log_bytes=$(wc -c < strix_scan.log)
+          echo "exit_code=$rc" >> "$GITHUB_OUTPUT"
+          echo "elapsed=$elapsed" >> "$GITHUB_OUTPUT"
+          echo "log_bytes=$log_bytes" >> "$GITHUB_OUTPUT"
+          echo "strix exited $rc after ${elapsed}s, wrote ${log_bytes} bytes of log"
+
+          # A non-zero exit AND no findings file means there is no evidence the
+          # scan produced a result at all -- fail, rather than triaging an empty
+          # run as a clean bill of health. A findings-free scan that exits 0 is
+          # still allowed through (#194): that is a real result.
+          if [ "$rc" -ne 0 ] && [ ! -f "${run_dirs[-1]}vulnerabilities.json" ]; then
+            echo "::error::strix exited $rc after ${elapsed}s and wrote no vulnerabilities.json -- no evidence the scan produced a result, so it is not reported as a clean scan. strix_scan.log in the private artifact repo has the reason; it is never printed here because it can contain scan content."
+            exit 1
+          fi
+
       - name: Dedupe against open issues + build triage report
         if: steps.scan.outcome == 'success'
         env:
@@ -218,6 +304,8 @@ jobs:
           cat step-summary.md >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "Model: \`${{ steps.model.outputs.model }}\` · scan mode: \`${{ inputs.scan_mode || 'deep' }}\`" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "strix exit code \`${{ steps.scan.outputs.exit_code }}\` · ran for ${{ steps.scan.outputs.elapsed }}s · wrote ${{ steps.scan.outputs.log_bytes }} bytes of log. A \`deep\` scan that finishes in a couple of minutes did not really run, whatever it reports." >> "$GITHUB_STEP_SUMMARY"
           if [ '${{ steps.model.outputs.fallback }}' = 'true' ]; then
             echo "" >> "$GITHUB_STEP_SUMMARY"
             echo "> ⚠️ **Fallback model** — the intended model is not served by the gateway, so this scan's coverage is not comparable to previous ones." >> "$GITHUB_STEP_SUMMARY"
@@ -232,7 +320,12 @@ jobs:
         # way to debug a startup failure, and it must never be printed to
         # this (public) job log -- the private repo is the one place it's
         # safe to look, whether or not the scan actually produced findings.
-        if: always()
+        #
+        # Skipped only when the preflight itself failed. There is nothing to
+        # push in that case -- the scan never started -- and attempting it would
+        # stack a second, misleading error on top of the one that already said
+        # exactly what to fix.
+        if: always() && steps.archive.outcome == 'success'
         env:
           SECURITY_SCAN_PAT: ${{ secrets.SECURITY_SCAN_PAT }}
           # Repo name lives in a secret, not in this public source.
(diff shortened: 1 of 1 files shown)


nhuelstng:
The token rotation worked, the preflight passed, and the archived log answered the open question from the issue: **why strix dies on startup.**

```
agents.sandbox.errors.LocalDirReadError: failed to read local dir artifact:
  /home/runner/work/oh-my-agentic-coder/oh-my-agentic-coder/target
```

## Root cause: strix refuses to copy a tree containing a symlink

`strix --target` hands the checkout to openai-agents, which walks it with `O_NOFOLLOW` and aborts on the first symlink it meets — `openai_agents-0.14.6/agents/sandbox/entries/artifacts.py:283`:

```python
if stat.S_ISLNK(entry_stat.st_mode):
    raise LocalDirReadError(
        src=src_root,
        context={"reason": "symlink_not_supported", "child": rel_child.as_posix()},
    )
```

The reason and the offending filename are in `context`, but `LocalDirReadError.__init__` builds its message from the path alone (`agents/sandbox/errors.py:603`):

```python
message=f"failed to read local dir artifact: {src}"
```

so neither ever reaches the log. That is why four weeks of runs looked like an unexplained early exit.

**The symlink is `CLAUDE.md -> AGENTS.md`**, the repo's only one:

```
$ git ls-tree -r HEAD | awk '$1=="120000"{print $4}'
CLAUDE.md
$ git log --format='%h %ad %s' --date=short -- CLAUDE.md
8948e24 2026-07-16 docs: add CLAUDE.md as symlink to AGENTS.md
```

Added **2026-07-16 — two days after this workflow was added** (`6ff2326`, 2026-07-14). So the security scan has never once run. The ~105 seconds were the docker image pull, and every run reported `Tokens 0` / `Cost $0.0000`.

Reproduced deterministically by replaying that exact walk over a clean checkout:

```
$ git worktree add --detach wt HEAD && python3 walk_repro.py wt
RAISED: failed to read local dir artifact: .../wt
    context: reason=symlink_not_supported child=CLAUDE.md
```

## Fix

Stage a dereferenced copy and scan that, rather than delete the symlink. `CLAUDE.md` exists so tools looking for either name find the same file; a pentest agent should see the content those tools actually read; and this way the next symlink somebody commits does not silently kill the scan again.

```
--- strix -n --target "$PWD/target"
+++ strix -n --target "$PWD/staging/target"
```

Unresolvable links are rejected up front, naming them, because `cp -L` would otherwise fail with a message about the copy rather than the repo. The check uses `find -type l` + `readlink -e`, **not** `find -xtype l` — the latter stats the target, so a symlink *loop* makes `find` exit non-zero having printed nothing to stdout, which would abort the job with no message at all. Found that the hard way while testing.

Same walk, before and after staging:

```
real repo (one symlink: CLAUDE.md)
  staged a symlink-free copy of the checkout; flattened: CLAUDE.md
  exit=0   strix walk: walked OK: 442 files
plus a dangling link
  ::error::...these links do not resolve (broken, or a loop): dangling
  exit=1
plus a symlink loop
  ::error::...these links do not resolve (broken, or a loop): loop-b loop-a
  exit=1
no symlinks at all
  staged a symlink-free copy of the checkout; flattened: none
  exit=0
```

`CLAUDE.md`'s content survives the dereference, and the flattened link is the *only* difference between the two trees (`diff -r --no-dereference`).

## Note on the guard added earlier in this PR

This failure is exactly what the exit-code guard was for: strix exits non-zero **and** writes no `vulnerabilities.json`, while still leaving a run directory behind (`Essential scan data saved to: .../strix_runs/target_4f80`). Before this PR that combination passed as a clean scan; now it fails. So even if the symlink fix regresses, the scan cannot go green without having produced a result.

nhuelstng:
the scan works but github does not like actions running for more than 3h :D

nhuelstng:
**Scope note** — this PR has grown one step beyond #281 as written, because fixing the startup failure exposed the next problem behind it. Flagging it here rather than silently widening the branch.

With the symlink fix in place, the first run that actually scanned ([34258819205](https://github.com/TNG/oh-my-agentic-coder/actions/runs/34258819205)) reached **22 findings and archived none of them**. It ran `17:43:38Z → 20:44:39Z` — exactly `timeout-minutes: 180` — and GitHub cancelled the job.

Two causes, and the second is the one worth reading twice:

1. **The job's own ceiling ended the scan.** A cancellation is the worst way for a long scan to end: the step never finishes, and everything downstream is skipped or reads outputs it never got to write.
2. **`Push full report` can only find the run directory through `steps.scan.outputs.run_dir`, which is written *after* strix returns.** It was empty, so the step copied nothing — while reporting success — with the entire run directory sitting on the runner's disk, about to be torn down with it. Any step that is the last thing between hours of work and an empty commit should rediscover what it needs rather than trust an output the failure path never sets.

Both are fixed on this branch rather than a parallel one, since a second branch on this file would only conflict — same reasoning as the model pin. Partial scans are now archived and triaged as `PARTIAL` instead of being filed next to runs that died in 105 seconds.

Happy to split this into its own issue if you would rather keep the history one-problem-per-issue; the commit is self-contained (`c9c7001`) and would move cleanly.

One thing this does **not** settle: whether a scan reaches the end inside the new 300-minute budget. The 3h run was still working when it was cut off, so 300 is an estimate, not a measurement. The next real run is what tells us.

mwtng (approved):
Can you streamline the comments before merging? My pair-reviewing agent found already several inconsistencies there that make the review harder and might mislead future agents. Feel free to also shorten preexisting ones. 

IMHO most of those comments belong rather in an issue description or the why-section of the PR instead. (Probably true for any "historic comments" of the kind "in the past, xyz") 
Especially since agents tend to add comments instead of adjusting them, this just increases review complexity, time and token costs where the code itself already says what it is doing (and the git history what was done before).

Anyways, the code changes check out.

nhuelstng:
Comments streamlined in `2e02ff0` — 208 lines of comment down to 155, no code changes.

**What went:** the history. Which runs failed and on which dates, what the code used to say (`|| true`, `timeout-minutes: 180`, the inline model probe), how long the symlink bug had been there. That belongs in the PR description and the git log, so it is now only there.

**What stayed:** the reasons a reader needs to touch the code safely. Why `find -type l` and not `-xtype l`, why `-s INT` and not TERM, why `openai` is pinned, why the log is never printed on a public runner.

**Three things the comments had wrong, fixed while in there:**

1. The job timeout comment said a cancelled job skips the archive step. It does not — that step runs under `always()`, it just reads outputs the scan step never wrote. That distinction is the whole reason the run-directory rediscovery exists.
2. The strix invocation still said there was "no cost cap beyond this job's timeout-minutes", written before the scan got its own `timeout` budget in the same PR.
3. The preflight error said "Not starting a three-hour scan" with the budget now at 300 minutes. Now "multi-hour" — the one non-comment line in the diff.

Verified the diff is comments-only by stripping comment lines from both revisions and comparing; the only code difference is that error string.

PR #286, opened 2026-09-13
feat(nix): package omac with a flake

### LINKED ISSUE #285: Provide packaging for nix (NixOS)
### Duplicate check

### Context / Summary

In order to use this software on NixOS or other systems using the nix package manager, it needs to be packaged for nix. When it is not available upstream (https://github.com/nix-community/nixpkgs), offering a nix flake is a common way of providing this functionality.

### Problem / What

Package for nix is missing

### Suggested fix / Ask

Create a nix flake to create a reusable package for nix users.

### Non-goals

Contribute the package upstream at nixpkgs

### Evidence / Environment

_No response_

mithodin:
**Issue:** Closes #285

## What
- Provide release-pinned omac packages for Linux x86-64/ARM64 and macOS ARM64 through flake-utils.
- Add a NixOS module with configurable harness packages available through the omac wrapper, not installed system-wide.
- Automate release metadata updates, permanent `nix-<version>` tags, and packaging PRs to main.

## Why
NixOS users need reproducible installation without requiring upstream contributors to update a Nix dependency hash whenever Go dependencies change.

## How
- Keep the version, exact release source commit, source hash, and vendorHash directly in `flake.nix`; currently pin v0.9.0. Preserve the locked nixpkgs and flake-utils inputs.
- Expose `omac.agents.<name>.enable` and `.package`; selected harnesses are prepended to the omac wrapper's PATH while existing PATH fallback remains available.
- After GoReleaser publishes a `v*` release, calculate the new hashes and require native Nix checks on all three supported systems before publishing.
- Atomically push a packaging branch and permanent `nix-*` tag, then open a repository-template PR. Never push directly to main, force-push, or replace existing tags; retries preserve the tagged commit.
- Document installation, maintenance, publication, and recovery in the Quick Start and contributor guide.

## Verification
- Passed `EXPECTED_VERSION=0.9.0 sh scripts/flake_test.sh` (all-system evaluation, native Linux package/smoke build, binary version check, and module evaluation).
- Passed `nix flake check`; `nix run .#omac -- version` reports `omac 0.9.0`.
- Passed the real updater round trip: `python3 scripts/update-nix-release.py v0.9.0 a44213177e298ecdf2a5161004185bcc98000579`, leaving the pinned metadata unchanged.
- Passed `python3 scripts/update-nix-release_test.py` (5 tests), `python3 scripts/nix-release-workflow_test.py` (29 tests using temporary local Git remotes), and the existing release-policy test.
- Passed workflow shell syntax checks, actionlint on the new workflow, actionlint without ShellCheck on the modified existing workflows, and ShellCheck on the publication/smoke scripts.
- Passed `npm run build` in `website/` and `git diff --check`.
- Live GitHub publication and ARM/macOS native builds were not run locally; the workflow gates publication on those native builds. The full Go suite previously encountered environment-specific failures; this change does not modify Go source.

## Follow-up
- Configure `NIX_RELEASE_TOKEN` with Contents and Pull requests read/write access, and protect `nix-*` tags against updates/deletion. No main-branch bypass is required.
- Run the workflow's dry-run mode before the first automated Nix publication. Existing application tags remain unchanged; consumers use the permanent Nix tags once published.

Generated with OpenCode.


mwtng (changes requested):
Seems fine with me in general, as it does not seem to affect the other installation methods.

However, could you move the Nix section below the main supported OSes in the quick-start.md file (at least below MacOS and Debian/Ubuntu, maybe just before the "From source" section)?
Also mark it as "(experimental)" in the title please, since we are not regularly testing the compatibility with that OS.
(And while you are at it, please also add that experimental tag to the "Mise" installation as well, since we are not regularly testing with that one as well.)

